PR #96, opened 2014-10-22
Remove the default value for the integer-simple flag.

acw:
Using a default seems to require users to explicitly override the flag if they
want to enable it. Leaving it off allows Cabal to automatically figure out the
proper value of the flag, which is handy, especially when text is at the bottom
of a long dependency chain.


scravy:
This change breaks the build of several packages that depend on text. integer-simple is not on hackage and so cabal build will fail. Cabal does not correctly infer the value for this.

PR #111, opened 2015-01-31
Implement repeat, cycle, iterate for lazy Texts

RyanGlScott:
In which I attempt to define `repeat`, `cycle`, and `iterate` similarly to how [lazy `ByteString`](http://hackage.haskell.org/package/bytestring-0.10.4.1/docs/Data-ByteString-Lazy.html#g:10) does. I have no idea if this can be fused or inlined more, so I will leave that to someone with more knowledge of high-performance Haskell than me.

### DIFF
--- a/Data/Text/Lazy.hs
+++ b/Data/Text/Lazy.hs
@@ -125,7 +125,10 @@ module Data.Text.Lazy
     , mapAccumR
 
     -- ** Generation and unfolding
+    , repeat
     , replicate
+    , cycle
+    , iterate
     , unfoldr
     , unfoldrN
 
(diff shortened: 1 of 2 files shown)


Data/Text/Lazy.hs
```diff
@@ -929,6 +939,21 @@ replicate n t
                  | otherwise = t : rep (i+1)
 {-# INLINE [1] replicate #-}
 
+-- | 'cycle' ties a finite 'Text' into a circular one, or equivalently,
```
meiersi: I'd suggest to add the non-empty precondition to the documentation as follows.

'cycle' ties a finite, non-empty 'Text' into a circular one ....

meiersi:
I've reviewed the code and the implementation itself looks fine, apart from the small documentation improvement that I noted.

What is missing from my point of view is that you also replicate the tests from the bytestring library: see <https://github.com/haskell/bytestring/blob/cb85a5360bc540c88b3ae1886d07c741bec3cdaa/tests/Properties.hs>

RyanGlScott:
Thanks for the feedback. I changed the documentation and added three new tests: `tl_repeat`, `tl_cycle`, and `tl_iterate`. Again, I don't know if this is considered comprehensive enough, so feel free to point out anything sketchy I did.

Also, I had to change the implementation of `iterate`, since the previous one was not sufficiently lazy for testing purposes.

meiersi:
Looks good to me. @bos what do you think?

bos:
Thanks!

PR #117, opened 2015-04-29
Fix bug that caused encoding Int to loop forever.

andrewthad:
This is @meteficha's suggestion that fixes https://github.com/bos/text/issues/114 and https://github.com/bos/aeson/issues/246. I have tested it locally.


bos:
Thanks!

PR #119, opened 2015-05-04
Add feature parity Data instance Text and Text.Lazy

christiaanb:
I was making a `Data.Data` instance for a data type that has a `Data.Text.Lazy.Text` field, and noticed it was failing on `toConstr`, even though this worked for `Data.Text.Text`.

I then saw there was a disparity between the two instances, and assumed this was an oversight when the instance for `Data.Text.Text` was updated. So I went ahead and updated the `Data.Data` instance for `Data.Text.Lazy.Text` so that it behaves the same way as the instance for `Data.Text.Text`.


bos:
Nice, thanks.

PR #124, opened 2015-05-15
Don't leak space when passing the undecoded portion to the next iteration

tomjaguarpaw:
Fixes https://github.com/bos/text/issues/123


bos:
Nice!

PR #128, opened 2015-07-21
Add Data.Text.Lazy.Builder.fromShow

zyla:


bos:
Thanks for your patch. This doesn't seem valuable enough to be worth adding.

PR #129, opened 2015-07-27
Show: Activate literal rewrite rules in simplifier phase 1

bgamari:
Previously these rewrite rules were quite fragile as they could compete with the `unpack` rule defined in `GHC.Base`. This was tickled by 7.10.2, which happened to rewrite the LHS of rules with other rules (which is itself a bug in GHC). This caused the literal rules to fail to fire, which then resulted in long compilation times and poor code generation.

Unfortunately it took a couple of attempts to get this right. The first attempt moved the rules forward in compilation to phase 2; this did not resolve the issue but this fact was hidden by a silly testing mistake. I then added a proper test of the literal rules to prevent future mistakes and found that moving the rules to simplifier phase 1 resolved the issue. 

See [GHC Trac #10528](https://ghc.haskell.org/trac/ghc/ticket/10528) for further discussion.

### DIFF
--- a/Data/Text/Show.hs
+++ b/Data/Text/Show.hs
@@ -50,19 +50,19 @@ unpackCString# :: Addr# -> Text
 unpackCString# addr# = unstream (S.streamCString# addr#)
 {-# NOINLINE unpackCString# #-}
 
-{-# RULES "TEXT literal" [2] forall a.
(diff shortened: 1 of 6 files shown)


bos:
Thanks for sorting this out, Ben!

PR #132, opened 2015-08-12
Add `PrintfArg` instances for `Text` (strict and lazy).

aloiscochard:
Hi,

I would like to add instances for `PrintfArg`, I need them and that seems quite useful to have it directly in the library.

I'm not sure if those instances should be guarded behind a conditional checking on GHC version?

Let me know what you think

Thanks in advance


bos:
Nice, thanks!

Issue #89, opened 2014-09-16
Add takeWhileEnd


Fuuzetsu:
Is there a reason why there's no `takeWhileEnd`? There's `takeEnd` and `dropWhileEnd` so I don't see why not.

jwaldmann:
+1

bos:
Go ahead and add it.

PR #135, opened 2015-10-07
Implement takeWhileEnd

paulcapron:
This commit adds a `takeWhileEnd` function, both for strict and lazy Text, closing #89.
Appropriate QuickCheck tests are also added.

I’ve stick as much as possible to the style of the existing `takeWhile` / `dropWhileEnd` / etc.

I am no low-level Haskell expert though, and they are a couple of things I dumbly copy without fully
understanding what they do:

- in Lazy.hs, I use `{-# INLINE takeWhileEnd #-}` and not `{-# INLINE [1] takeWhileEnd #-}`.
This is similar to the existing `{-# INLINE dropWhileEnd #-}` but unlike most inline declarations which use the `[1]`; I don’t know what is the correct declaration to use here.
- in Text.hs, I have added `{-# RULES …}` by mere copy/pasting; I don’t fully understand this RULES thing.
- in Lazy.hs, I did NOT add `{-# RULES …}` because `dropWhileEnd` don’t have them. I’m not sure if lazy `takeWhileEnd` could have RULES, unlike `dropWhileEnd`, or not. I don’t understand if `dropWhileEnd` don’t have RULES because none are applicable for it or if they are just missing.

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -129,6 +129,7 @@ module Data.Text
     , drop
     , dropEnd
     , takeWhile
+    , takeWhileEnd
     , dropWhile
     , dropWhileEnd
     , dropAround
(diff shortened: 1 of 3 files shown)


bos:
Lovely, thanks!

PR #136, opened 2015-10-12
Add examples to Data.Text transformations

alexliew:
Added some examples for Data.Text transformation functions. Ref: https://github.com/MaxGabriel/Haskell-Documentation/issues/11.

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -598,7 +598,9 @@ compareLength t n = S.compareLengthI (stream t) n
 -- | /O(n)/ 'map' @f@ @t@ is the 'Text' obtained by applying @f@ to
 -- each element of @t@.
 --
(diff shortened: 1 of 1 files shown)


Data/Text.hs
```diff
@@ -596,8 +596,13 @@ compareLength t n = S.compareLengthI (stream t) n
 -- -----------------------------------------------------------------------------
 -- * Transformations
 -- | /O(n)/ 'map' @f@ @t@ is the 'Text' obtained by applying @f@ to
--- each element of @t@.  Subject to fusion.  Performs replacement on
--- invalid scalar values.
+-- each element of @t@.
+--
+-- Example:
+--
+-- > map (\c -> if c == '.' then '!' else c) $ pack "I am not angry. Not at all." ==> "I am not angry! Not at all!"
```
bos: This line is way longer than 80 characters. If you could please fold it, I'd appreciate it.

PR #144, opened 2016-01-16
Text.Array: Use unsafeFreezeByteArray# in place of unsafeCoerce#

bgamari:
It shouldn't hurt to use the type-safe variant here. Moreover, this
unsafeCoerce# usage was responsible for STG lint errors
(GHC Trac #8114).


bos:
Nice, thanks!

PR #146, opened 2016-02-18
add Generics instance to Data.Text

kuk0:
we want Generics for everything (otherwise we can't derive Generics for any data types which contain these without creating orphan instances)

the end-goal is to have all of these: https://gist.github.com/kuk0/92fff1c91c526ffb0a9e
and be able to derive instances for e.g. TextShow or Text.PrettyPrint.Generic.Pretty

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -5,6 +5,7 @@
 #endif
 #if __GLASGOW_HASKELL__ >= 708
 {-# LANGUAGE TypeFamilies #-}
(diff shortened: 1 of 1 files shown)


bos:
This is interesting but both entirely magical and undocumented and has no tests. I thus have no idea how to evaluate whether this code makes any sense.

RyanGlScott:
I don't think a `Generic` instance would be appropriate for the same reasons laid out in https://github.com/haskell/vector/pull/109#issuecomment-188911383.

PR #147, opened 2016-03-30
[WIP] Various rule-robustness issues

bgamari:
The fragile-rule warnings introduced in GHC 8.0 have pointed out a few more places where the rules in `text` may not fire,

 * `Data.Text.length` may be inlined before the `compareLength` rules have a change to fire
 * the `compareLength` rules match against class ops
 * `Data.Text.Lazy.Builder.Internal.flush` may be inlined before the `flush/flush` rule has a chance to fire

I still need to test this against GHC <8.0 and check for performance regressions.


bos:
Nice. I'll chew through this soon.

PR #151, opened 2016-05-14
Optimize unstream loop (3x better on micro benchmark)

harendra-kumar:
Affected macro benchmarks' improvement range from 2%  to 34%. More details are available in the commit message. 

I have also added corresponding micro benchmarks. Also added strictness annotation for `replicateCharI` args.

This change was motivated by a perf measurement for pure Haskell unicode normalization code that I am working on. While comparing the performance with text-icu normalization I found that just stream/unstream with no other processing in between was taking more time than what text-icu was taking for normalizing English text. Now, with this change stream/unstream is performing better than that.


bos:
Really nice work, thank you!

PR #155, opened 2016-07-02
Depend on bytestring-builder for API backwards compatibility

RyanGlScott:
This was previously proposed in #67, which was ultimately rejected due to concerns about requiring the use of `bytestring-builder` in a Haskell Platform release that shipped GHC 7.8. That Haskell Platform has since been released, so I don't think it's a concern anymore.

This PR ensures that `encodeUtf8Builder` is always defined, even on older versions of GHC, by depending on the `bytestring-builder` package if a particularly old version of `bytestring` is being used. This dependency shouldn't be a concern going forward, since even if, say, `text` were to become a GHC dependency, building `text` as part of the GHC build process wouldn't incur a `bytestring-builder` dependency.

A real-world example of where this would be useful is in https://github.com/bos/aeson/pull/426#issuecomment-225358139, where we had to work around the fact that `text` doesn't export `encodeUtf8Builder` when `text` is built against an old version of `bytestring`.

/cc @lpsmith

### DIFF
--- a/Data/Text/Encoding.hs
+++ b/Data/Text/Encoding.hs
@@ -55,13 +55,9 @@ module Data.Text.Encoding
     , encodeUtf32LE
     , encodeUtf32BE
 
-#if MIN_VERSION_bytestring(0,10,4)
     -- * Encoding Text using ByteString Builders
-    -- | /Note/ that these functions are only available if built against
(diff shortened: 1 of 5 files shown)


bos:
Thanks, Ryan!

PR #161, opened 2016-08-17
Added a stack.yaml file with stack init

NorfairKing:


mgsloan:
If a smaller file is desired, just `resolver: lts-6.12` will also suffice

bos:
I've never used stack at all, so I'm afraid you'll have to explain to me (preferably in the commit description) what this is even for. Also, all the bloaty auto-generated commentary is not useful.

PR #166, opened 2016-10-11
Fix documentation for  Data.Text.Read hexadecimal

HuwCampbell:
Embarrassingly trivial PR here... docs were saying decimal digit instead of hexadecimal digit.


bos:
Thanks!

PR #169, opened 2016-12-06
Added additional test for toTitle

alskipp:
Ensures that the first letter of each word is not lowercase after
`toTitle` transform. Doesn’t check to ensure that it’s uppercase, as
not all alphabets contain uppercase and lowercase letters.


bos:
Cheers!

bos:
lololololol 831626dcc7a779ccdff6866bbb7d22326950e0eb

alskipp:
Ouch! The joy of human languages and Unicode : )

PR #175, opened 2017-04-21
Add catchable failure decoders for more encodings:

mrijkeboer:
- decodeUtf16LE'
- decodeUtf16BE'
- decodeUtf32LE'
- decodeUtf32BE'

### DIFF
--- a/Data/Text/Encoding.hs
+++ b/Data/Text/Encoding.hs
@@ -34,6 +34,10 @@ module Data.Text.Encoding
 
     -- ** Catchable failure
     , decodeUtf8'
+    , decodeUtf16LE'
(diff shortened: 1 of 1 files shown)


bos:
These seem like good ideas, but I have two observations.

The names aren't exactly self-documenting :-)

The implementations seem a bit unnatural. Basically if you're round-tripping through `unsafePerformIO . try . evaluate`, maybe the underlying primitives have the wrong return types?

mrijkeboer:
The code is based on the already provided [decodeUtf8'](https://github.com/haskell/text/blob/master/Data/Text/Encoding.hs#L307)  function. I tried to keep it as consistent with the existing code as possible.

Issue #176, opened 2017-05-03
takeEnd is wrong for surrogate pairs


kuk0:
for example:
```
> I.showText "fo𐐷"
"Text [102,111,55297,56375] 0 4"
> I.showText $ T.takeEnd 1 "fo𐐷"
"Text [56375] 3 1"               -- instead of "Text [55297,56375] 2 2"
> I.showText $ T.takeEnd 2 "fo𐐷"
"Text [111,55297,56375] 1 3"     -- correct
[] λ> I.showText $ T.takeEnd 2 "f𐐷o"
"Text [56375,111] 2 2"           -- instead of "Text [55297,56375,111] 1 3"
```

the base case  `| m <= 1    = i` is wrong in `iterNEnd` since the invariant is: `i` points to the **last** `Word16` of the code point, not the first one

kuk0:
this is bad because then e.g.:
```
> T.encodeUtf8 $ T.takeEnd 2 "f𐐷o"
"\237\176\183o"

> T.decodeUtf8 . T.encodeUtf8 $ T.takeEnd 2 "f𐐷o"
*** Exception: Cannot decode byte '\xb0': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```

kuk0:
`dropEnd` has a similar problem:
```
> I.showText $ T.dropEnd 2 "f𐐷o"
"Text [102,55297] 0 2"
```

cf that with `drop`, which works fine:
```
> I.showText  $ T.drop 1 "f𐐷o"
"Text [55297,56375,111] 1 3"
> I.showText  $ T.drop 2 "f𐐷o"
"Text [111] 3 1"
```

i looked at `Test.QuickCheck.Arbitrary` and found part of the problem:
```
instance Arbitrary Char where
  arbitrary = chr `fmap` oneof [choose (0,127), choose (0,255)]
```
i.e. it seems that most of the tests for Data.Text (if not all) just check the behaviour for ASCII...

PR #177, opened 2017-05-03
fix takeEnd and dropEnd - issue #176

kuk0:
see #176 

a separate issue is that Data.Text is not tested with non-ascii chars
to test this diff i created 

```
newtype MyString = MyString String deriving Show

instance Arbitrary MyString where
  arbitrary = MyString <$> oneof [listOf (chr `fmap` oneof [choose (0,127), choose (65536, 1114111)])]
  shrink (MyString s) = map MyString $ shrink s
```

and used it to generate random strings...
the `t_takeEnd` test showed broken with this change and this diff fixed it


bos:
Thanks!

PR #185, opened 2017-06-17
Data.Text.Lazy.cons is O(1) rather than O(n)

steshaw:
It looks like `Data.Text.Lazy.cons` is O(1). This is a documentation change only.


bos:
Thanks!

Issue #183, opened 2017-05-25
Add tshow function


mightybyte:
In multiple different projects that I have worked on I've encountered variations on the following function:

```haskell
tshow :: Show a => a -> Text
tshow = T.pack . show
```

It would be great if the text package could add this to eliminate the redundancy and confusion.

RyanGlScott:
A variant of this has been proposed before (and rejected) in https://github.com/bos/text/issues/106.

mightybyte:
That's actually a different thing.  The `text-show` package gives you a standalone type class specifically for the purpose of converting to a Builder.  I'm specifically talking about the existing `Show` class.  That package also has more dependencies than text, and is therefore not an acceptable answer for what I want.

ryantrinkle:
Perhaps it would make sense as `Data.Text.show`?  In the spirit of many other functions that are defined in `Data.Text` and generally imported qualified (e.g. `null`, `map`, `reverse`, etc.), this would be a direct analogy of `Prelude.show`.

mightybyte:
I don't have much of an opinion what the name is.  I'm just interested in squashing the unnecessary proliferation and fragmentation of pointless reimplementations this function.

alexeyzab:
Hi there! Is this issue still relevant? If so, would any of the maintainers be interested in providing some mentorship/general help on resolving this so that the newcomers and first-time contributors would have an easier time?

I would like to add this as one of the issues for the upcoming [Haskell Weekly's](https://github.com/haskellweekly/haskellweekly.github.io) Call for Participation section. See discussion in [#75](https://github.com/haskellweekly/haskellweekly.github.io/issues/75).

These are the guidelines we'd like to stick to in the future:
* Ensure that your project has at least one open-source licence. (we've decided to not define the term "open-source" and it is left to your own interpretation).
* Ensure that the issue tracker for your project is publicly accessible.
* Create a new issue in your issue tracker and clearly describe the task. Also mention the difficulty level (easy/medium/hard/tedious), either as a tag/label or somewhere in the title/description.
* If you have specific requirements for contributors (e.g., copyright waiver), it must be mentioned in the description of the task (preferably with a link to CONTRIBUTING.md).

Thank you!

mightybyte:
This is definitely still relevant, and it's a very easy issue to fix.  The main question at this point is maintainer buy-in.

taktoa:
Personally, I've written this function upwards of 20 times. +1

codygman:
[...]

PR #186, opened 2017-07-14
Adding a tshow function (closing #183)

Eugleo:
A simple alies that allows one to use Show instances with Text.


bos:
Thanks for submitting this, but as mentioned on the original issue, I'm not going to take it.

PR #198, opened 2017-09-08
Fix #197

### LINKED ISSUE #197: Incorrect result for `(== 1) . length . filter (== ',')` depending on -O level
@raichoo reports a nasty bug related to `text` (it's still unclear what exactly is going on here, although I suspect `text`'s RULE framework to be the culprit -- I hope so at least, as the alternative would mean that several GHC releases out there may generate wrong code):

```hs
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE BangPatterns #-}
module Main where

import Prelude hiding (length, filter)
import Data.Text

currencyParser :: Text -> Bool
currencyParser x = cond == 1
  where
    cond = length fltr
    fltr = filter (== ',') x

main :: IO ()
main = print (currencyParser "0,00")
```

when compiling this with `-O0` or running in GHCi, the proper `True` output is generated. However, when compiling with `-O1` the wrong `False` output is generated.

bgamari:
This fixes a variety of size hint bugs in `text`'s fusion framework. These could manifest in a variety of issues similar to #197.

### DIFF
--- a/Data/Text/Internal/Fusion/Common.hs
+++ b/Data/Text/Internal/Fusion/Common.hs
@@ -857,7 +857,8 @@ indexI (Stream next s0 _len) n0
 -- returns a stream containing those characters that satisfy the
 -- predicate.
 filter :: (Char -> Bool) -> Stream Char -> Stream Char
-filter p (Stream next0 s0 len) = Stream next s0 len -- HINT maybe too high
+filter p (Stream next0 s0 len) =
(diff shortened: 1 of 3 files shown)


bgamari:
Thanks to reminders from @hvr and @nomeata I'm going to pick this up again.

bgamari:
Arg, yet again I regret opening a PR from my `master` branch.

bgamari:
This has been superceded by #200.

PR #200, opened 2017-10-09
Fix usage of size hints

bgamari:
This fixes a variety of size hint bugs in `text`'s fusion framework. It seems these were introduced when the `Size` type was extended to track both upper and lower bounds in f4fc30cea2a41bd26b6c4445843ba1b3f7d0dde8. These could manifest in a variety of issues similar to #197.

### DIFF
--- a/Data/Text/Internal/Fusion/Common.hs
+++ b/Data/Text/Internal/Fusion/Common.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns, MagicHash, Rank2Types #-}
+{-# LANGUAGE PatternGuards, BangPatterns, MagicHash, Rank2Types #-}
 -- |
 -- Module      : Data.Text.Internal.Fusion.Common
 -- Copyright   : (c) Bryan O'Sullivan 2009, 2012
@@ -117,7 +117,7 @@ import GHC.Prim (Addr#, chr#, indexCharOffAddr#, ord#)
 import GHC.Types (Char(..), Int(..))
 
 singleton :: Char -> Stream Char
-singleton c = Stream next False 1
+singleton c = Stream next False (codePointsSize 1)
     where next False = Yield c True
           next True  = Done
 {-# INLINE [0] singleton #-}
@@ -175,7 +175,7 @@ data C s = C0 !s
 
 -- | /O(n)/ Adds a character to the front of a Stream Char.
 cons :: Char -> Stream Char -> Stream Char
-cons !w (Stream next0 s0 len) = Stream next (C1 s0) (len+1)
+cons !w (Stream next0 s0 len) = Stream next (C1 s0) (len + codePointsSize 1)
     where
       next (C1 s) = Yield w (C0 s)
       next (C0 s) = case next0 s of
@@ -189,7 +189,7 @@ data Snoc a = N
 
 -- | /O(n)/ Adds a character to the end of a stream.
 snoc :: Stream Char -> Char -> Stream Char
-snoc (Stream next0 xs0 len) w = Stream next (J xs0) (len+1)
+snoc (Stream next0 xs0 len) w = Stream next (J xs0) (len + codePointsSize 1)
   where
     next (J xs) = case next0 xs of
       Done        -> Yield w N
@@ -237,7 +237,7 @@ uncons :: Stream Char -> Maybe (Char, Stream Char)
 uncons (Stream next s0 len) = loop_uncons s0
     where
       loop_uncons !s = case next s of
-                         Yield x s1 -> Just (x, Stream next s1 (len-1))
+                         Yield x s1 -> Just (x, Stream next s1 (len - codePointsSize 1))
                          Skip s'    -> loop_uncons s'
                          Done       -> Nothing
 {-# INLINE [0] uncons #-}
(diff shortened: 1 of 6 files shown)


Data/Text/Internal/Fusion/Common.hs
```diff
@@ -318,10 +318,13 @@ lengthI (Stream next s0 _len) = loop_length 0 s0
...
 compareLengthI :: Integral a => Stream Char -> a -> Ordering
-compareLengthI (Stream next s0 len) n =
-    case compareSize len (fromIntegral n) of
-      Just o  -> o
-      Nothing -> loop_cmp 0 s0
+compareLengthI (Stream next s0 len) n
+    -- Note that @len@ tracks code units whereas we want to compare the length
+    -- in code points. Specifically, a stream with hint @len@ may consist of
+    -- anywhere from @len/2@ to @len@ code points.
+  | Just LT <- compareSize (2*len) (fromIntegral n) = LT
```
Mistuke: Is this one correct? Since `n` is in code points, and a code point can be 2 code units, won't
`len=6` and `n=1` give `LT` here while it should be `GT`? I would have thought that the original compare just lacked a call to `codePointsSize` over `n`?

Data/Text/Internal/Fusion/Common.hs
```diff
@@ -699,7 +702,7 @@ replicateI n (Stream next0 s0 len) =
 -- (a,b), in which case, a is the next Char in the string, and b is
 -- the seed value for further production.
 unfoldr :: (a -> Maybe (Char,a)) -> a -> Stream Char
-unfoldr f s0 = Stream next s0 1 -- HINT maybe too low
+unfoldr f s0 = Stream next s0 unknownSize
```
Mistuke: Don't you drop `s` in ever stream, so can't `unknownSize` be `charSize s`?

Data/Text/Internal/Fusion/Common.hs
```diff
@@ -395,7 +398,7 @@ intersperse c (Stream next0 s0 len) = Stream next (I1 s0) len
 
 caseConvert :: (forall s. Char -> s -> Step (CC s) Char)
             -> Stream Char -> Stream Char
-caseConvert remap (Stream next0 s0 len) = Stream next (CC s0 '\0' '\0') len
+caseConvert remap (Stream next0 s0 _) = Stream next (CC s0 '\0' '\0') unknownSize
```
Mistuke: hmm unless I'm misunderstanding something, can't this be `len + maxSize $ charSize s0` ?

Data/Text/Internal/Fusion/Common.hs
bgamari: I'm not sure I follow. Can you rephrase?

Data/Text/Internal/Fusion/Common.hs
bgamari: Oh dear, yes, I believe you are right.

Data/Text/Internal/Fusion/Common.hs
Mistuke: What I mean is, in each step of the stream you consume one input character don't you? So according to the comment

     -- The function takes the element and returns
     -- Nothing if it is done producing the stream or returns Just
     -- (a,b), in which case, a is the next Char in the string, and b is 
    -- the seed value for further production. 

So isn't the initial length the number of code points of the seed?

bgamari: Ahh, I see.
> So isn't the initial length the number of code points of the seed?

Not quite, since the user might take and double the size of each code point (turning one-unit code-point into a two-unit code point). Consequently, I think we can set the upper bound at `2*hint`.

bgamari: Let's try this again.

> What I mean is, in each step of the stream you consume one input character don't you?

This is where you lose me. What input stream are we talking about here? `unfoldr` produces a stream by unfolding starting from an initial value, successively applying the given function until it returns `Nothing`. Consequently I'm not sure how we could possibly know how long the stream will be; it depends entirely on the function.

Data/Text/Internal/Fusion/Common.hs
bgamari: Okay, so I looked at the call sites and indeed they *do* produce only one code point per input code point although there is of course nothing in the type of `caseConvert` that requires this be the case. However, given that `caseConvert` isn't exported, I suppose we can just state this as an invariant.

bgamari: On more careful reading it seems that the point I made above actually wasn't quite right: `remap` can produce up to three codepoints for every input code-point.

Mistuke: Ah ok, thanks for the explanation!

Data/Text/Internal/Fusion/Common.hs
Mistuke: Ah, there's my misunderstanding. I had thought it only applied it once to get the next value. In which case I figured you always know the max amount of code points the next would produce. Thanks for clearing that up!

hvr:
Looks good to me; Thanks everyone!

bgamari:
Can we get a release with this soon?

PR #204, opened 2017-10-29
use environments to lazily initialize benchmarks

kuk0:
if we only need to run a few benchmarks,
e.g. `text-benchmarks Pure/drop/Text+tiny`
it's wasteful to initialize all of them
(and read all the test data in memory - all
the test data is a few hundreds of MB, but
we read it in several tests, as different types,
and most seriously, as String - which turns this
into GBs of memory)

### DIFF
--- a/benchmarks/haskell/Benchmarks.hs
+++ b/benchmarks/haskell/Benchmarks.hs
@@ -5,7 +5,7 @@ module Main
     ( main
     ) where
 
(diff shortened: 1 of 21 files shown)


kuk0:
in fact, i would like to either split or completely delete the Strings from the benchmarks; just reading 100MB of text as a String is slow and i'm not really convinced the String benchmarks are useful

hvr:
Thanks!

PR #205, opened 2017-10-29
fix github links in README

kuk0:
https://github.com/bos/text doesn't exist anymore
it was moved to https://github.com/haskell/text


hvr:
Thanks I meant to update the links but forgot; just for the record, the links do exist insofar as they will be HTTP redirected

PR #206, opened 2017-10-29
improve Makefile for getting test-data

kuk0:
add a target for git and update docs

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,3 +1,4 @@
+VCS = hg
 count = 1000
 
 all: coverage literal-rule-test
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -15,6 +15,10 @@ text-test-data:
 	hg clone https://bitbucket.org/bos/text-test-data
 	$(MAKE) -C text-test-data
 
+text-test-data-git:
```
hvr: There's a difference to the `text-test-data` target though: the `text-test-data` target actually creates the filename it's named as, whereas `text-test-data-git` doesn't result in a folder `text-test-data-git` being created.

kuk0: @hvr: true, thanks; i've updated the commit

hvr:
thanks!

PR #208, opened 2017-11-03
faster compareText

kuk0:
we don't need to iterate over whole codepoints and decode them into Char's
it is ok to go word by word; this works also for 2-word codepoints
this saves ~10% of time, which is important for sorting/Map Text
benchmark:

```
before:

benchmarking Programs/Sort/Text
time                 591.7 ms   (577.7 ms .. 600.9 ms)
mean                 594.2 ms   (591.8 ms .. 595.8 ms)
std dev              2.352 ms   (0.0 s .. 2.713 ms)
variance introduced by outliers: 19% (moderately inflated)

after:

benchmarking Programs/Sort/Text
time                 531.2 ms   (527.8 ms .. 533.7 ms)
mean                 533.2 ms   (532.3 ms .. 533.7 ms)
std dev              855.6 μs   (0.0 s .. 857.6 μs)
variance introduced by outliers: 19% (moderately inflated)
```

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -410,17 +410,20 @@ textDataType = mkDataType "Data.Text.Text" [packConstr]
 
 -- | /O(n)/ Compare two 'Text' values lexicographically.
 compareText :: Text -> Text -> Ordering
-compareText ta@(Text _arrA _offA lenA) tb@(Text _arrB _offB lenB)
-    | lenA == 0 && lenB == 0 = EQ
-    | otherwise              = go 0 0
+compareText (Text arrA offA lenA) (Text arrB offB lenB)
+    | lenA == 0 || lenB == 0 = compare lenA lenB
+    | otherwise              = go offA offB
   where
+    ea = offA + lenA
+    eb = offB + lenB
     go !i !j
-        | i >= lenA || j >= lenB = compare lenA lenB
-        | a < b                  = LT
-        | a > b                  = GT
-        | otherwise              = go (i+di) (j+dj)
-      where Iter a di = iter ta i
-            Iter b dj = iter tb j
+        | i >= ea || j >= eb = compare lenA lenB
+        | a < b              = LT
+        | a > b              = GT
+        | otherwise          = go (i+1) (j+1)
+      where
+      a = A.unsafeIndex arrA i
+      b = A.unsafeIndex arrB j
 
 -- -----------------------------------------------------------------------------
 -- * Conversion to/from 'Text'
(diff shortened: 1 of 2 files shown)


kuk0:
the improvement is actually 25-50% (when comparing two equal strings - to decide whether x<x we need to compare all characters)
note that `Programs/Sort` above benchmarks the whole program, which consists of reading the data, splitting them into lines and sorting the lines and i only improved sorting

```
benchmarking Pure/compareText/Text+tiny
time                 82.79 ns   (82.76 ns .. 82.83 ns)
time                 60.84 ns   (60.79 ns .. 60.93 ns)
benchmarking Pure/compareText/Text+ascii-small
time                 588.1 μs   (588.0 μs .. 588.3 μs)
time                 293.9 μs   (293.8 μs .. 294.0 μs)
benchmarking Pure/compareText/Text+ascii
time                 510.9 ms   (510.5 ms .. 511.3 ms)
time                 257.1 ms   (256.4 ms .. 257.6 ms)
benchmarking Pure/compareText/Text+english
time                 34.14 ms   (34.13 ms .. 34.15 ms)
time                 17.19 ms   (17.18 ms .. 17.20 ms)
benchmarking Pure/compareText/Text+russian
time                 54.49 μs   (54.48 μs .. 54.51 μs)
time                 27.26 μs   (27.26 μs .. 27.26 μs)
benchmarking Pure/compareText/Text+japanese
time                 37.88 μs   (37.87 μs .. 37.89 μs)
time                 18.82 μs   (18.81 μs .. 18.82 μs)
```

hvr:
@kuk0

> it is ok to go word by word; this works also for 2-word codepoints

I'm not convinced this is actually sound

For instance, consider the code-points `U+E001` and `U+10001`

Their UTF16 encoding into 16 bit words would be

```
0xe001
0xd800 0xdc01
```

respectively; and yet if you compare them lexicographically as 16-bit words, you'd end up determining that `U+E001 > U+10001` which is obviously not correct.

Am I missing something?

llelf:
@hvr you are not missing anything, it is wrong.

However there *actually exists* a way to do it, using the additional lookup table, see Unicode core specification, p5.17.   But I’m not quite sure it will be worth it.

This is one another case where UTF-8 is better (because doing it in UTF-8 is straightforward) and why UTF-16 in `Text` was a huge mistake.

hvr:
@llelf yeah, I suspect the performance improvement will likely be rather modest for the added complexity of using the lookup tables...

Lysxia:
Fixing this seems like a rather involved affair, so it seems safe to close this.

PR #210, opened 2017-12-09
Document that Data.Text.IO throws on decoding failure

rightfold:
This not being documented has bitten me in the past, causing downtime.


hvr:
@rightfold thank you! Could you maybe also document which type of exception is thrown in that case?

Issue #211, opened 2017-12-28
Restrictive upper bound on QuickCheck 2.10.1 hides potential segfault


snoyberg:
QuickCheck-2.10.1 is out of bounds for:
- [ ] text-1.2.3.0 (>=2.7 && < 2.10). No maintainer. @haskell. Used by: test-suite

hvr:
This is intentional: the test-suite isn't yet compatible with QC 2.10's semantics, hence why I set the upper bound.

snoyberg:
I see two test failures with the newer QuickCheck:

```
    errors:
      t_utf8_err: [Failed]
*** Failed! (after 6 tests): 
Exception:
  Data.Text.Array.unsafeWrite: bounds error, offset 4, length 4
  CallStack (from HasCallStack):
    error, called at ./Data/Text/Array.hs:163:37 in main:Data.Text.Array
Solo
Replace
"\240\129\140\154"
(used seed -4447769247167586605)
```

```
      t_infix_concat: [Failed]
*** Failed! (after 52 tests and 16 shrinks): 
Exception:
  Data.Text.Array.unsafeIndex: bounds error, offset 2, length 2
  CallStack (from HasCallStack):
    error, called at ./Data/Text/Array.hs:155:36 in main:Data.Text.Array
"\128"
"a"
""
Just '\428388'
(used seed -7715035259973790205)
```

What I'm concerned about in these results is that this looks like it's uncovering a legitimate bug in the library, not a mismatch with new semantic in QuickCheck. Given the importance of text in the ecosystem, this is something worth looking into. (I'm doing so now, but if you have some insights, that would be great.)

hvr:
It's got to do with `decodeUtf8With` having an undocumented invariant/assumption; but if that assumption doesn't hold you'd have to significantly over-allocate by a factor of 2 iirc.

snoyberg:
So to confirm: there's a known, undocumented aspect of `decodeUtf8With` which can lead to segfault, and this behavior is triggered by the new QuickCheck library but not the older one?

snoyberg:
I've opened up PR #212 as a workaround for this.

PR #212, opened 2017-12-29
Prevent potential segfault in decodeUtf8With #211

snoyberg:
I considered documenting the corner case in the Haddocks for `decodeUtf8With`, but thought the corner case may be more confusing than not, and didn't want it to block this PR. If you'd prefer something be added, I'll be happy to do so.

### DIFF
--- a/Data/Text/Encoding.hs
+++ b/Data/Text/Encoding.hs
@@ -73,7 +73,7 @@ import Data.ByteString.Internal as B hiding (c2w)
 import Data.Text.Encoding.Error (OnDecodeError, UnicodeException, strictDecode)
 import Data.Text.Internal (Text(..), safe, text)
 import Data.Text.Internal.Private (runText)
-import Data.Text.Internal.Unsafe.Char (ord, unsafeWrite)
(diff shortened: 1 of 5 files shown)


snoyberg:
Another approach that would work here would be to have an `unsafeWrite1` or similar function, and throw an exception if the value is above `'\ffff'`. Since that conditional is already being checked in `unsafeWrite`, that should be no runtime performance impact for the common case.

snoyberg:
Build failures appear to be the same as those on `master`.

snoyberg:
I don't have a problem with (2), I considered that approach also. But to be
sure we're on the same page: how does the current approach break the
internal invariants? I'm not intimately familiar with the invariants of
Text, I'd like to know where I screwed up. I thought my implementation was
actually (3).

hvr:
> how does the current approach break the internal invariants?

It allows invalid UTF-16 code units to be written into the `Text` buffer

snoyberg:
I've pushed an update which implements (2) instead.

bgamari:
What is the status of this?

PR #224, opened 2018-06-10
Add configuration for Appveyor CI

berdario:
This [currently fails](https://ci.appveyor.com/project/berdario/text/build/1.0.3)

```
  hGetContents_crash: [Failed]
ERROR: DeleteFile ".\\crashy4118467.txt": permission denied (The process cannot access the file because it is being used by another process.)
```

Not sure if this fails also on other Windows environments (I don't have a Windows machine here at the moment)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,19 @@
+build: off
+
+platform:
+- x64
+
+before_build:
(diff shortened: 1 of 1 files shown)


Lysxia:
Given that we are now switching to Github Actions it doesn't seem like we need an Appveyor config.

PR #226, opened 2018-07-31
Fix termination condition for file reads.

### LINKED ISSUE #223: Certain content is truncated upon read, on Windows
@birbacher found an issue when reading certain files on Windows.

We tried with bare `hGetContents`, we changed the buffering, changed the compilation options `-O0`, `-O1`, etc... but haven't yet found a difference.

Adding or removing a single character before the truncation point is enough for the test case to pass, but otherwise it fails every single time.

It seems to only truncate on a line ending after a multiple of 4k (that's why we attempted to change the buffering for it).

I created a small repo with a test case:

https://github.com/berdario/text-bug-repro

https://github.com/berdario/text-bug-repro/blob/master/src/Lib.hs `readTheSame` is supposed to yield `True`

And this is the file in question: https://github.com/berdario/text-bug-repro/blob/master/testfile

You can see the failure live here: https://ci.appveyor.com/project/berdario/text-bug-repro/build/1.0.4

Mistuke:
This fixes #223 

The termination condition of `Text`'s `hGetContents` was relying on the `EOF` marker being replaced by a single `\r`. This would go wrong when the only decode-able character left in the byte buffer is a single `\r`.

`Text` would then think it reached the `EOF` and would stop reading. This changes the termination condition to just keep reading until it actually reaches an `EOF`.

### DIFF
--- a/Data/Text/IO.hs
+++ b/Data/Text/IO.hs
@@ -95,14 +95,22 @@ writeFile p = withFile p WriteMode . flip hPutStr
 appendFile :: FilePath -> Text -> IO ()
 appendFile p = withFile p AppendMode . flip hPutStr
 
-catchError :: String -> Handle -> Handle__ -> IOError -> IO Text
+catchError :: String -> Handle -> Handle__ -> IOError -> IO (Text, Bool)
 catchError caller h Handle__{..} err
     | isEOFError err = do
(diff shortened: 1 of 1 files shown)


hvr:
Thanks!

PR #229, opened 2018-08-22
Fix #227

### LINKED ISSUE #227: Error when calling Text.take with a negative 'n' value.
For example:

```
main :: IO ()
main = print $ T.length $ T.filter isLetter $ T.take (-3) "Hello! How are you doing today?"
```

throws the following error:

```
Data.Text.Array.new: size overflow
CallStack (from HasCallStack):
  error, called at libraries/text/Data/Text/Array.hs:132:20 in text-1.2.3.0:Data.Text.Array
```

The problem only seems to happen with O1 or higher, so it is likely related to the fusion case.

Also, the problem seems to have be introduced somewhere after text-1.2.2.2.

bgamari:
This fixes two bugs:

 * The size in `caseConvert` was incorrectly (albeit subtly) associated, resulting in a much large size that overflowed the machine integer
 * The `take` and `drop` fusion combinators failed to clamp negative counts, resulting in underflows.

In addition it does a fair bit of spring cleaning, adding GHC 8.6.1 to the `Tested-With` configuration list, dropping a redundant `bug-reports` field, and regenerating `travis.yml`.


.travis.yml
```diff
@@ -25,87 +28,129 @@ before_cache:
...
 matrix:
   include:
-    - compiler: "ghc-7.0.4"
-    # env: TEST=--disable-tests BENCH=--disable-benchmarks
-      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-7.0.4], sources: [hvr-ghc]}}
-    - compiler: "ghc-7.2.2"
+    - compiler: "ghc-8.6.1"
+      env: GHCHEAD=true
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.6.1], sources: [hvr-ghc]}}
+    # Fails due to GHC #15436
```
RyanGlScott: When you say "fails", is it the entire library that fails to build? Just the tests?

If it's the former, we shouldn't list GHC 8.4.3 under `tested-with`. If it's the latter, maybe we should uncomment this and simply enable `TEST=--disable-tests`.

bgamari: Just the tests.

> If it's the former, we shouldn't list GHC 8.4.3 under tested-with. If it's the latter, maybe we should uncomment this and simply enable TEST=--disable-tests.

Sounds good to me.

hvr:
@bgamari I need to start writing down all the issues w/ GHC 8.4.3 I'd like to see resolved in a GHC 8.4.4 point rls...

bgamari:
How is this looking?

hvr:
@bgamari good :-) 

But I'll test it locally a bit more

PR #230, opened 2018-08-22
Fix #221 and #227

### LINKED ISSUE #221: toLower crashes GHC for large inputs
It seems to want to allocate huge amounts of memory. It iven semms to overflow:

```
Prelude> import qualified Data.Text as T
Prelude T> :set -XOverloadedStrings 
Prelude T> T.toLower (T.replicate 200000 "0")
"<interactive>: internal error: Unable to commit -1604321280 bytes of memory
    (GHC version 8.2.2 for x86_64_unknown_linux)
    Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
```

bgamari:
This fixes two bugs:

 * The size in `caseConvert` was incorrectly (albeit subtly) associated, resulting in a much large size that overflowed the machine integer
 * The `take` and `drop` fusion combinators failed to clamp negative counts, resulting in underflows.

In addition it does a fair bit of spring cleaning, adding GHC 8.6.1 to the `Tested-With` configuration list, dropping a redundant `bug-reports` field, and regenerating `travis.yml`.


hvr:
How does this relate to #229?

bgamari:
It's the same work; originally I was going to split it into two PRs but I decided against this.

PR #232, opened 2018-08-30
Add Lift instances for Text.

quasicomputational:
These use the existing `Data` instances, and have similar trade-offs.

I've also bumped the version number appropriately, but I don't know if that'd be better done later (e.g., by the release manager near the time of release).

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -6,6 +6,17 @@
 #if __GLASGOW_HASKELL__ >= 708
 {-# LANGUAGE TypeFamilies #-}
 #endif
+-- Using TemplateHaskell in text unconditionally is unacceptable, as
+-- it's a GHC boot library. TemplateHaskellQuotes was added in 8.0, so
+-- this would seem to be a problem. However, GHC's policy of only
+-- needing to be able to compile itself from the last few releases
+-- allows us to use full-fat TH on older versions, while using THQ for
+-- GHC versions that may be used for bootstrapping.
+#if __GLASGOW_HASKELL__ >= 800
+{-# LANGUAGE TemplateHaskellQuotes #-}
+#else
+{-# LANGUAGE TemplateHaskell #-}
+#endif
 
 -- |
 -- Module      : Data.Text
@@ -244,6 +255,8 @@ import GHC.Base (eqInt, neInt, gtInt, geInt, ltInt, leInt)
 #if __GLASGOW_HASKELL__ >= 708
 import qualified GHC.Exts as Exts
 #endif
+import qualified Language.Haskell.TH.Lib as TH
+import Language.Haskell.TH.Syntax (Lift, lift)
 #if MIN_VERSION_base(4,7,0)
 import Text.Printf (PrintfArg, formatArg, formatString)
 #endif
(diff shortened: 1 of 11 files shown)


hvr:
I'd prefer to delay this to the next release; as 1.2.3.1 is the one supposed to ship w/ GHC 8.6.1 and I'd like to avoid last minute changes  (especially API additions) w/o having had the time to think them through

quasicomputational:
Yeah, no rush on this. This has just annoyed me one too many times and I figured I'd send a patch.

hvr:
I'm very sympathetic to his PR; @quasicomputational could you come up with a couple simple test-cases that execute the new codepaths?

quasicomputational:
Good call on the tests. I could have sworn I checked it was working with GHCi, but, having written the tests, it seems like the `Data` approach was broken. Just goes to show!

Along the way, I got bitten by a cabal-install bug, haskell/cabal#5623. That's why the TH-requiring tests are in their own little package.

Data/Text.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns, CPP, MagicHash, Rank2Types, UnboxedTuples #-}
```
hvr: Requiring full TH to build `text` is not acceptable; however, `TemplateHaskellQuotes` would be acceptable

quasicomputational: This is awkward. `TemplateHaskellQuotes` is only since 8.0, and I'm drawing a blank on how to write this without quote syntax: `mkName` and `lookupValueName` operate at the use site, which is unsuitable (what if users don't import `Data.Text`? What if it's been renamed?).

Might have to put this on ice until 7.10 and lower are out of the support window.

hvr: Well, if you can make it work for GHC >= 8 with THQ, we can conditionally enable old-style TH for GHC < 8 . We just need to make sure that recentish GHCs don't require to have a bytecode interpreter in order to build the core library `text`

quasicomputational:
I'm seeing a proptest failure in `t_compareLength` locally, but I didn't touch that code so I'm hoping it'll pass in Travis and it's just some transitory weirdness.

text.cabal
```diff
@@ -147,7 +147,8 @@ library
     base       >= 4.2 && < 5,
     binary,
     deepseq    >= 1.1.0.0,
-    ghc-prim   >= 0.2
+    ghc-prim   >= 0.2,
+    template-haskell
```
brandon-leapyear: Hi, just curious, was any thought put into adding this extra dependency? Now, packages which use `text` will now need to build/link `template-haskell`. Is this not that a big enough issue for most people?

hvr: Note that `template-haskell` is considered a non-reinstallable package; i.e. it's only built once when GHC itself is build. Also note this doesn't require a GHC with TH engine either (i.e. `text` still works perfectly well with a GHC that cannot run TH). But why do you consider the need to link against `template-haskell` a problem?

brandon-leapyear: If a Haskell executable statically links their dependencies, wouldn't it become non-trivially bigger having to statically link `template-haskell`, even if they aren't using the `Lift` instance for `Text`?

PR #234, opened 2018-09-19
fix text-tests.cabal

kuk0:
`text-tests` doesn't build because `Tests.IO` has module name `Main`
`Tests.IO` has a separate build target `text-tests-stdio` and it shouldn't be here
  
```
        File name does not match module name:
        Saw: ‘Main’
        Expected: ‘Tests.IO’

    4 | module Main
      |        ^^^^
```


hvr:
Good catch; thanks!

Issue #202, opened 2017-10-17
Fusion not working as advertised


nomeata:
The documentation of `Data.Text` states

> Most of the functions in this module are subject to fusion, meaning that a pipeline of such functions will usually allocate at most one Text value.
> 
> As an example, consider the following pipeline:
> ```
> import Data.Text as T
> import Data.Text.Encoding as E
> import Data.ByteString (ByteString)
>  
> countChars :: ByteString -> Int
> countChars = T.length . T.toUpper . E.decodeUtf8
> ```
> From the type signatures involved, this looks like it should allocate one ByteString value, and two Text values. However, when a module is compiled with optimisation enabled under GHC, the two intermediate Text values will be optimised away, and the function will be compiled down to a single loop over the source ByteString.
> 
> Functions that can be fused by the compiler are documented with the phrase "Subject to fusion".

but (with GHC-8.0 or 8.2 and text-1.2.2.2) this is not true: Looking at `-ddump-simpl` I clearly see a `Text` value constructed of which the length is calculated.

I assume that originally, the example from the introduction did fuse properly, and then it broke…

Do you have a test suite that somehow tests the fusion promises of `text`? I stumbled over this because I am thinking about a proper, structured way of testing such non-functional properties of code in test suites, based on first successes with [using  `ghc-proofs` for that in `generic-lens`](https://github.com/kcsongor/generic-lens/commit/3fddfc5b9fde0904e452a1866515fd493b914c67#diff-3ae00181b173238649ff56154b9d01e4R43), and maybe `text` would be a good second case study.

nomeata:
I did a more comprehensive test of testing properties of the `text` library, and the [test harness that I created](https://github.com/nomeata/inspection-testing/tree/master/examples/text-api) spits out this information:
```
Data.Text:
Total functions (95)
Untestable functions (32)
Fuses (40): pack, decodeUtf8, singleton, unfoldr, unfoldrN, cons, snoc, append, tail, init, map, intersperse, toCaseFold, toLower, toUpper, toTitle, justifyLeft, scanl, take, drop, takeWhile, dropWhile, filter, unpack, head, last, null, compareLength, foldl, foldl', foldl1, foldl1', foldr, foldr1, any, all, maximum, minimum, findIndex, isPrefixOf
Does not fuse (but should) (11): unpackCString#, reverse, scanl1, scanr1, takeWhileEnd, dropWhileEnd, dropAround, strip, stripStart, stripEnd, length
[...]

PR #241, opened 2018-11-16
Address some of the fusion issues in #202

nomeata:
in most cases, just updating the docs. More interesting stuff for `takeWhileEnd` and `length`, see the corresponding commits. See #202.


nomeata:
Is this mergeable?

hvr:
@nomeata yes, it's mergeable... thanks :-)

PR #246, opened 2019-01-16
Fix -incomplete-uni-patterns

### LINKED ISSUE #245: Two incomplete-uni-patterns
There are 2 `incomplete-uni-patterns` in this library. Since this warning will soon be enabled by -Wall, we should either a) use `-Wno-...` , or b) case+error.

The offender:
```
d:ds' = map i2d (if ei > 0 then is' else 0:is')
```

in `Data.Text.Lazy.Builder.RealFloat` : [L92](https://github.com/haskell/text/blob/master/Data/Text/Lazy/Builder/RealFloat.hs#L92) and [L120](https://github.com/haskell/text/blob/master/Data/Text/Lazy/Builder/RealFloat.hs#L120).

ckoparkar:
Fixes #245.


ckoparkar:
@hvr please take a look :) I used case+error instead of `-Wno`.

hvr:
@ckoparkar thanks!

PR #261, opened 2019-05-26
Add paddedDecimal, for zero-padding.

quasicomputational:
This doesn't currently specialise as well as `decimal`, but it should
be easy to give it an analogous structure if that would help. In the
absence of benchmarks, I've left it in the simpler form.


quasicomputational:
Also, I've speculatively put this as being since 1.2.4. If events happen on the way to getting this merged and released, that may need to be updated to keep it from being a lie.

quasicomputational:
Ping? Is there anything I need to do before this is ready to merge?

PR #262, opened 2019-06-12
Data.Text.Internal.Functions: remove

vaibhavsagar:
Data.List has provided `intersperse` with the desired performance characteristics since at least base 4.8.0.0.


vaibhavsagar:
It looks like this change happened between [`base-4.3.1.0`](http://hackage.haskell.org/package/base-4.3.1.0/docs/src/Data-List.html#intersperse) and [`base-4.4.0.0`](http://hackage.haskell.org/package/base-4.4.0.0/docs/src/Data-List.html#intersperse).

hvr:
So this sounds like this will need more work in order to make sure this doesn't break operational semantics with `base-4.3.*` (aka GHC 7.0)?

vaibhavsagar:
Yes, that's right. Alternatively we could drop GHC 7.0 support, but it doesn't sound like that's going to happen.

PR #263, opened 2019-07-11
Add ForeignFunctionInterface flag for compilation

ch3pjw:
I don't know if this is the right things to do, but I was trying to get this repository to compile by running
```
stack init
stack build
```
and I got errors of the form
```
/home/paul/src/haskell/text/tests/../Data/Text/Encoding.hs:522:9: error:
    parse error on input ‘import’
    |
522 | foreign import ccall unsafe "_hs_text_decode_utf8" c_decode_utf8
    |         ^^^^^^
```
from a couple of modules.

Adding the `ForeignFunctionInterface` extension to the `LANGAUGE` pragma at the top of those files made it work for me, just not sure it's the right thing for everyone's build configuration.


hvr:
This sounds like something that was already resolved in Git; at least I can't reproduce this with cabal. Generally I strongly recommend to use cabal if you intend to contribute/work on `text` as that's the baseline I use myself and where I can help you best if any problems should arise.

PR #264, opened 2019-09-07
Replace flag 'integer-simple' with 'integer-gmp'

ch1bo:
Also simplify CPP switches as nothing 'integer-simple' related is used.

### DIFF
--- a/Data/Text/Lazy/Builder/Int.hs
+++ b/Data/Text/Lazy/Builder/Int.hs
@@ -36,7 +36,7 @@ import Prelude hiding ((<>))
 #endif
 
 #ifdef  __GLASGOW_HASKELL__
-# if defined(INTEGER_GMP)
+# if defined(MIN_VERSION_integer_gmp)
 import GHC.Integer.GMP.Internals (Integer(S#))
 # endif
 #endif
@@ -189,7 +189,7 @@ hexDigit n
 data T = T !Integer !Int
 
 integer :: Int -> Integer -> Builder
-#ifdef INTEGER_GMP
+#if defined(MIN_VERSION_integer_gmp)
 integer 10 (S# i#) = decimal (I# i#)
 integer 16 (S# i#) = hexadecimal (I# i#)
 #endif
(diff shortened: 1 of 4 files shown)


text.cabal
```diff
@@ -161,10 +160,7 @@ library
     ghc-options: -Werror
     cpp-options: -DASSERTS
 
-  if flag(integer-simple)
-    cpp-options: -DINTEGER_SIMPLE
-    build-depends: integer-simple >= 0.1 && < 0.5
-  else
+  if flag(integer-gmp)
     cpp-options: -DINTEGER_GMP
```
hvr: We can actually simplify this even more; i.e. we can get rid of `-DINTEGER_GMP` as it's kinda redundant; use

```c
#if defined(MIN_VERSION_integer_gmp)
/* assume GMP is available */
#elif defined(MIN_VERSION_integer_simple)
/* assume integer-simple is used */
#else
# error unsupported configuration
#endif
```

instead  to detect which one of those packages is in scope (and assume that

Data/Text/Lazy/Builder/Int.hs
```diff
@@ -38,19 +38,9 @@ import Prelude hiding ((<>))
 #ifdef  __GLASGOW_HASKELL__
 # if defined(INTEGER_GMP)
 import GHC.Integer.GMP.Internals (Integer(S#))
-# elif defined(INTEGER_SIMPLE)
-import GHC.Integer
-# else
-# error "You need to use either GMP or integer-simple."
```
hvr: where did this error go to?

text.cabal
ch1bo: Good idea - changed in https://github.com/haskell/text/pull/264/commits/34736352382a16bb20a6083b06cb48aec8f9aaf8

Data/Text/Lazy/Builder/Int.hs
ch1bo: Not required as we only need to know whether gmp-specific things shall be included (the `S#` constructor)

Lysxia:
Some of the simplifications here have taken place in #283, but not all, and there are now some conflicts. @ch1bo would you be interested in updating this PR? If yes I'll take care of reviewing this.

Bodigrim:
I do not find this change desirable for the reasons similar to https://github.com/haskell/bytestring/pull/182#issuecomment-626396007: 
* `integer-simple` is an automatic flag, users should not be setting it manually, so its name is of little importance, 
* for GHC 9.0+ the flag is irrelevant because of `ghc-bignum`.

ch1bo:
Yeah, I suppose this is redundant now. Will close the PR.

PR #267, opened 2019-09-24
Bump template-haskell upper bound

bgamari:
For GHC `master`.

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -161,7 +161,7 @@ library
     binary           >= 0.5 && < 0.9,
     deepseq          >= 1.1 && < 1.5,
     ghc-prim         >= 0.2 && < 0.6,
-    template-haskell >= 2.5 && < 2.16
+    template-haskell >= 2.5 && < 2.17
 
   if flag(bytestring-builder)
     build-depends: bytestring         >= 0.9    && < 0.10.4,


bgamari:
@hvr, ping.

bgamari:
@hvr, ping.

bgamari:
@hvr, what can we do to get this in?

hvr:
@bgamari are the API changes in 2.16 really benign?

RyanGlScott:
I'm not @bgamari, but...

> @bgamari are the API changes in 2.16 really benign?

...I believe so, judging from the [changelog](https://gitlab.haskell.org/ghc/ghc/blob/4898df1cc25132dc9e2599d4fa4e1bbc9423cda5/libraries/template-haskell/changelog.md#21600-tba). The only parts of `template-haskell` that `text` uses are `Lift(lift)`, `appE`, `varE`, and `stringE`, and none of those have breaking changes in `template-haskell-2.16.0.0`.

However, there is one change that is worth noting: the `Lift` class has gained an additional method `liftTyped`, which does not have a default implementation in terms of `lift`. At the moment, `text` only implements `lift` in its `Lift` instances, which means that `text` emits these warnings when compiling with HEAD:

```
[28 of 44] Compiling Data.Text        ( Data/Text.hs, /tmp/ghc19598_0/ghc_52.o )

Data/Text.hs:431:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘Language.Haskell.TH.Syntax.liftTyped’
    • In the instance declaration for ‘Lift Text’
    |
431 | instance Lift Text where
    |          ^^^^^^^^^
```
```
[35 of 44] Compiling Data.Text.Lazy   ( Data/Text/Lazy.hs, /tmp/ghc19598_0/ghc_110.o )

Data/Text/Lazy.hs:416:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘Language.Haskell.TH.Syntax.liftTyped’
    • In the instance declaration for ‘Lift Text’
    |
416 | instance Lift Text where
    |          ^^^^^^^^^
```

I'll prepare a PR to fix this.

RyanGlScott:
See #269.

PR #269, opened 2019-11-08
Implement liftTyped in the Lift Text instances

RyanGlScott:
`template-haskell-2.16.0.0` adds a `liftTyped` method to `Lift`, which is like `lift` but returning `Q (TExp t)` instead of `Q Exp`. `liftTyped` does not have a default implementation in terms of `lift`, so it must be implemented explicitly in the `Lift Text` instances that `text` defines. The implementations I chose are based off of the general approach taken by `Language.Haskell.TH.Syntax`, such as in this example: https://gitlab.haskell.org/ghc/ghc/blob/4898df1cc25132dc9e2599d4fa4e1bbc9423cda5/libraries/template-haskell/Language/Haskell/TH/Syntax.hs#L716

### DIFF
--- a/Data/Text.hs
+++ b/Data/Text.hs
@@ -254,7 +254,7 @@ import GHC.Base (eqInt, neInt, gtInt, geInt, ltInt, leInt)
 import qualified GHC.Exts as Exts
 #endif
 import qualified Language.Haskell.TH.Lib as TH
-import Language.Haskell.TH.Syntax (Lift, lift)
+import qualified Language.Haskell.TH.Syntax as TH
 #if MIN_VERSION_base(4,7,0)
 import Text.Printf (PrintfArg, formatArg, formatString)
 #endif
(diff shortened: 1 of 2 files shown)


Data/Text.hs
```diff
@@ -254,7 +254,11 @@ import GHC.Base (eqInt, neInt, gtInt, geInt, ltInt, leInt)
 import qualified GHC.Exts as Exts
 #endif
 import qualified Language.Haskell.TH.Lib as TH
-import Language.Haskell.TH.Syntax (Lift, lift)
+import Language.Haskell.TH.Syntax ( Lift, lift
+#if MIN_VERSION_template_haskell(2,16,0)
```
hvr: I'm not a fan (anymore) of splitting up "statements" via CPP; this tends to confuse syntax-highlighters or indenters; please use "complete" statements (i.e. use a 2nd `import` statement guarded by CPP)

or even better yet; let's just use a qualified import `as TH` or similar, and avoid this CPP altogether in this place

Data/Text/Lazy.hs
```diff
@@ -248,7 +248,11 @@ import qualified GHC.Exts as Exts
 #endif
 import GHC.Prim (Addr#)
 import qualified Language.Haskell.TH.Lib as TH
-import Language.Haskell.TH.Syntax (Lift, lift)
+import Language.Haskell.TH.Syntax ( Lift, lift
```
hvr: here too

Data/Text.hs
RyanGlScott: Done.

hvr:
Thanks again!

PR #274, opened 2020-01-07
Add `elem` combinator to API

emilypi:
This PR adds the `elem` combinator to the API.

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -185,6 +185,7 @@ module Data.Text
     , filter
     , breakOnAll
     , find
+    , elem
     , partition
 
     -- , findSubstring
@@ -1502,6 +1503,13 @@ chunksOf k = go
 
 -------------------------------------------------------------------------------
 -- ** Searching with a predicate
+
+-- | /O(n)/ The 'elem' function takes a character and a 'Text', and
+-- returns 'True' if the element is found in the given 'Text', or
+-- 'False' otherwise.
+elem :: Char -> Text -> Bool
+elem c t = S.any (== c) (stream t)
+{-# INLINE elem #-}
 
 -- | /O(n)/ The 'find' function takes a predicate and a 'Text', and
 -- returns the first element matching the predicate, or 'Nothing' if
--- a/src/Data/Text/Lazy.hs
+++ b/src/Data/Text/Lazy.hs
@@ -191,6 +191,7 @@ module Data.Text.Lazy
     -- * Searching
     , filter
     , find
+    , elem
     , breakOnAll
     , partition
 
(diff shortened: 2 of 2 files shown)


Data/Text.hs
```diff
@@ -1488,6 +1490,13 @@ chunksOf k = go
 -------------------------------------------------------------------------------
 -- ** Searching with a predicate
 
+-- | /O(n)/ The 'elem' function takes a character and a 'Text', and
+-- returns 'True' if the element is found in the given 'Text', or
+-- 'False' otherwise.
+elem :: Char -> Text -> Bool
+elem c t = isJust (find (== c) t)
```
hvr, line 1497: Why not define this in terms of the `S.any` primitive that `Data.Text.any` is based upon?

```
any :: (Char -> Bool) -> Text -> Bool
any p t = S.any p (stream t)
{-# INLINE any #-}
```

?

hvr:
So I'm slightly on the fence about this; for one, `elem` is trivially definable in terms of `any`

```hs
elem c = any (==c)
```

and I'm not sure if this meets the https://wiki.haskell.org/Fairbairn_threshold standard.

Fwiw, there's appears to be lack of evidence that this is *needed* insofar as nobody seems to have missed having this trivial definition in `Data.Text` for over a decade...

However, there's precedent in e.g. `Data.ByteString.elem` and `Data.Vector.elem` which do provide this standard verb; so for the sake of API symmetry/uniformity among vector/list-ish data-structures, it'd make sense to add this (redundant) verb, as there might be some level of expectation of it existing.

Data/Text.hs
emilypi: Lack of knowledge of the `text` library is the reason. If that's the preferred way, let's do it.

emilypi:
> So I'm slightly on the fence about this; for one, elem is trivially definable in terms of any

My reasoning is precisely that this is the API that's defined for every other "stringlike" object. It passes the Fairbarn test on account of it's expected to be there, eminently useful (perhaps less so for text), and it homogenizes the APIs across libraries. You may object to it on the bounds that  it's a small addition, but it was *surprising* that it was not already.

Data/Text.hs
hvr: @emilypi let's do that then :-)

emilypi:
@hvr updated and it should be ready to go

emilypi:
@phadej too, if you're more active

emilypi:
@hvr rebased to master so we should be able to just merge with no problems

Lysxia:
This looks good to me (conclusion: we want this for uniformity with other collection-like APIs). Ping @haskell/text for another approval to merge this.

Bodigrim (approved):
Adding a test would be nice.

PR #293, opened 2020-07-16
Issue 277: make toCaseFold idempotent

phadej:
This PR currently only adds a test for #227. Expectedly it fails (when you run locally, run with `-p idemp -a 10000` to run just the new tests enough times to find a counterexample).

`toCaseFold` should be idempotent, but it doesn't seem to be.


src/Data/Text/Internal/Fusion/CaseMapping.hs
```diff
@@ -357,8 +359,196 @@ foldMapping '\x03f0' s = Yield '\x03ba' (CC s '\x0000' '\x0000')
...
+foldMapping '\x052c' s = Yield '\x052d' (CC s '\x0000' '\x0000')
+-- CYRILLIC CAPITAL LETTER EL WITH DESCENDER
+foldMapping '\x052e' s = Yield '\x052f' (CC s '\x0000' '\x0000')
 -- ARMENIAN SMALL LIGATURE ECH YIWN
 foldMapping '\x0587' s = Yield '\x0565' (CC s '\x0582' '\x0000')
+-- GEORGIAN CAPITAL LETTER YN
+foldMapping '\x10c7' s = Yield '\x2d27' (CC s '\x0000' '\x0000')
+-- GEORGIAN CAPITAL LETTER AEN
+foldMapping '\x10cd' s = Yield '\x2d2d' (CC s '\x0000' '\x0000')
+-- NOT FOLDED TO toLower '\x13a0'
```
phadej, line 380: This is an example of where `toLower` doesn't work. `'\x13a0` is a `Cherokee Letter A` which case folds to itself (i.e. stays uppercase!).

This file has also plenty of added letters with known name, i.e. in `CaseFolding.txt`. Those are capital letters for which there is folding to lowercase, but `toLower` doesn't always fold them. Adding them makes `toCaseFold` behave uniformly independently of `base` version.

phadej:
I won't fix the conflicts

PR #299, opened 2020-10-05
Introduce encodeASCII

ethercrow:
Someone might want to use that for performance. Preliminary benchmark of a specialized encodeASCII vs encodeUtf8:

```
benchmarking EncodeASCII/Text
time                 310.1 μs   (309.0 μs .. 311.5 μs)
mean                 312.3 μs   (311.1 μs .. 314.0 μs)
std dev              4.574 μs   (3.482 μs .. 5.920 μs)

benchmarking EncodeUtf8 (ASCII)/Text
time                 893.0 μs   (851.0 μs .. 923.3 μs)
                     0.942 R²   (0.861 R² .. 0.988 R²)
mean                 1.324 ms   (1.185 ms .. 1.535 ms)
std dev              686.6 μs   (472.2 μs .. 957.7 μs)
```

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -48,6 +48,7 @@ module Data.Text.Encoding
     , Decoding(..)
 
     -- * Encoding Text to ByteStrings
+    , encodeASCII
     , encodeUtf8
     , encodeUtf16LE
     , encodeUtf16BE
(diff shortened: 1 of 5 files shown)


src/Data/Text/Encoding.hs
```diff
@@ -443,6 +444,17 @@ encodeUtf8 (Text arr off len)
             memcpy ptr' ptr (fromIntegral utf8len)
             return (PS fp' 0 utf8len)
 
+-- | Encode text using ASCII encoding.
```
phadej, line 447: What happens when text has characters outside the ASCII range?

src/Data/Text/Lazy/Encoding.hs
```diff
@@ -161,6 +162,9 @@ encodeUtf8Builder :: Text -> B.Builder
 encodeUtf8Builder =
     foldrChunks (\c b -> TE.encodeUtf8Builder c `mappend` b) Data.Monoid.mempty
 
+encodeASCII :: Text -> B.ByteString
+encodeASCII = encodeUtf8
```
phadej, line 166: this is a lie.

src/Data/Text/Encoding.hs
ethercrow: Then the output is garbage. We can call it `unsafeEncodeASCII` to be clearer.

src/Data/Text/Lazy/Encoding.hs
ethercrow: Yes, this is just a placeholder, I didn't get to implementing the lazy version yet.

phadej: I usually liberally leave `TODO`s in the code, as hint for myself and others, but also as a reminder "what else I had to do". It's then hopefully easier to see whether everything I had in mind is done when I self-review the patch.

ethercrow:
Closing in favor of https://github.com/haskell/text/pull/300

PR #305, opened 2020-12-01
Use new primops names for converting numbers to and from the native size

Ericson2314:
This is a companion to https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4492.

I'm not sure why the second-to-top commit isn't already merged, as ghc HEAD relies on it.

CC @bgamari


phadej:
> I'm not sure why the second-to-top commit isn't already merged

The `[CmmSized] adjust`? I don't remember seeing PR for that. Ping @angerman?

---

Please don't add merge commits.

src/Data/Text/Array.hs
```diff
@@ -78,6 +78,17 @@ import GHC.ST (ST(..), runST)
 import GHC.Word (Word16(..))
 import Prelude hiding (length, read)
 
+#if MIN_VERSION_base(4,16,0)
+import GHC.Base ( word16ToWord#, wordToWord16# )
+#else
+import GHC.Prim (Word#)
+word16ToWord#, wordToWord16# :: Word# -> Word#
```
phadej, line 85: Let's put these into `Data.Text.Internal.PrimCompat`, next to `Data.Text.Internal.ByteStringCompat`.

angerman:
> > I'm not sure why the second-to-top commit isn't already merged
> 
> The `[CmmSized] adjust`? I don't remember seeing PR for that. Ping @angerman?

@bgamari took over and said he'd do the submodule bumps, as I was unsure what exactly needed doing.

bgamari:
Sigh, indeed, mea culpa. It looks like the linter failed to catch this as the `sized` commit was reachable via a "permanent" ref on `gitlab.haskell.org`. @phadej, I'm afraid that either we will need to accept a merge commit here or push a tag to ensure that the existing commit remains reachable to ensure that GHC's history doesn't break.

Submodules are just awful.

Ericson2314:
> @phadej, I'm afraid that either we will need to accept a merge commit here or push a tag to ensure that the existing commit remains reachable to ensure that GHC's history doesn't break.
>
> Submodules are just awful.

Yeah I think if we just dispense with any hope of linear history and try to reconcile `ghc-head` branches with upstream `master` on an asynchronous basis, it will be the least bad.

phadej:
The existing commit can be tagged with `oops-keep-me`.

phadej:
I converted this to draft. Please mark is as ready when it really is.

(Side comment, the history looks awful, please cleanup it before marking as ready fore review).

Ericson2314:
In https://mail.haskell.org/pipermail/libraries/2021-February/031186.html I begin discussing this PR on the libraries list.

phadej:
Fine. I'll merge this.

PR #306, opened 2020-12-05
Add examples to 'span' and 'break'

martijnbastiaan:
In case there are any doubts left after reading the docs :).


phadej:
Made small changes and merged in https://github.com/haskell/text/pull/308. Thanks.

PR #311, opened 2021-01-30
Use base-provided unsafe shifts

Ericson2314:
This makes the 9.2 prim changes less invasive. If we don't care about de-optimizing ancient GHCs we can remove this module altogether.


Ericson2314:
N.B. https://github.com/haskell/text/pull/305 is still needed to cope for the planned GHC 9.2 primop changes in some form, even if this PR is merged. But it will be smaller if this PR is merged. I will also try to adopt the tricks of https://github.com/haskell/containers/pull/760 and https://github.com/haskell/bytestring/pull/354/files for #305 to make code that is compatible both with and without the primop change plan without using CPP.

PR #318, opened 2021-03-15
Add Pull Request Template

parsonsmatt:
This PR adds a template for Pull Requests.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,14 @@
+Before submitting your PR, check that you've:
+
+- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
+- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock
+
+After submitting your PR:
+
+- [ ] Update the Changelog.md file with a link to your PR
+- [ ] Bumped the version number if there isn't an `(unreleased)` on the Changelog
+- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)
+
+<!---Thanks so much for contributing! :)
+
+_If these checkboxes don't apply to your PR, you can delete them_-->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,6 @@ Before submitting your PR, check that you've:
 After submitting your PR:
 
 - [ ] Update the Changelog.md file with a link to your PR
-- [ ] Bumped the version number if there isn't an `(unreleased)` on the Changelog
 - [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)
 
 <!---Thanks so much for contributing! :)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+Before submitting your PR, check that you've:
+
+- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
+- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock
```
parsonsmatt, line 4: I *really* want this. It's very nice to have `@since` annotations on code to establish lower bounds.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+
+After submitting your PR:
+
+- [ ] Update the Changelog.md file with a link to your PR
+- [ ] Bumped the version number if there isn't an `(unreleased)` on the Changelog
```
parsonsmatt, line 9: This may be unnecessary - `persistent` likes to release quickly, and this line makes it relatively easy to get things out fast.

phadej: `text` is not `persistent`. There is technically little value in releasing often as people only use versions bundled with GHC. (Stackage doesn't upgrade GHC bundled libs for example).

So `persistent` experience doesn't apply directly.

parsonsmatt: Of course - happy to remove the line :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+Before submitting your PR, check that you've:
+
+- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
+- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock
+
+After submitting your PR:
+
+- [ ] Update the Changelog.md file with a link to your PR
```
Bodigrim, line 8: I don't quite understand how it is supposed to work. I mean, a contributor fills out this form before submitting a PR, right?..

In my experience asking contributors to amend changelog is a source of merge conflicts, when there are several PR in flight. Contributors are also generally bad at following the same style of entries. IMO it's easier for maintainers to update changelog before release than chase everyone to edit and rebase.

Bodigrim (commented):
Great stuff!

.github/PULL_REQUEST_TEMPLATE.md
parsonsmatt: It works perfectly fine with `persistent`. It occasionally happens that you need to resolve a conflict, but it's easy enough.

Most contributors copy/paste the previous changelog item and fill in the relevant details. We could provide a common format in this file, to make it easier.

Bodigrim: OK, let's give it a try. But how is it supposed to work? Once a PR is created, should a contributor push an additional commit and edit message to cross out this item?

parsonsmatt: yup!

Bodigrim (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)
```
Lysxia, line 9: Github's UI takes care of this already.

Lysxia (commented):
I think the last point could be removed, and the comment at the end should be at the top.

.github/PULL_REQUEST_TEMPLATE.md
Lysxia: Contributors might not know what the next version number is going to be.

One solution is to preemptively create a changelog entry of X.Y.(Z+1).0 where X.Y.Z.P is the latest released version.

This also raises the questions of whether and how to manage incrementing the 4-th number, for releases without API additions. And also how to fit breaking changes in this workflow.

Lysxia:
@Bodigrim  Is there a policy for contributors to know upfront what the next version number is going to be? That seems a blocking thing to make the second item actionable. Should it be `X.Y.(Z+1).0` always, so by default we merge only PVP-minor-level changes, and breaking changes get merged once in a blue moon?

Bodigrim:
@Lysxia to be honest, my experience of burdening contributors with admin tasks is unsatisfactory. A new function could be targetting a major release, but end up being backported to a stable branch, etc. The practice of `bytestring` and `vector` is that a release manager goes through changes and adds `@since` pragmas where appropriate.

Boarders:
I agree with @Bodigrim, I also personally found it quite confusing when I was trying to add to a library and this issue came up.

PR #322, opened 2021-03-20
Add tests for elem

Lysxia:
Follow up to #274 

Also planning to

- [ ] add `@since` annotation (but what version?)
- [ ] add changelog entry (if we are to follow #318)


Bodigrim (approved):

Lysxia:
#318 is getting stale so let's not keep this blocked.

Bodigrim (approved):

PR #323, opened 2021-03-20
Modularize property tests

Lysxia:
The module `Tests.Properties` is **1.4K** lines long and takes **30 seconds** to build on my machine. Modularizing makes re-builds much less painful.


Lysxia:
Oh, compatibility with GHC 7 was broken while cleaning up imports for the test suite. Do we want to bite the bullet now, as [dropping GHC 7 is already in the plans](https://github.com/haskell/text/pull/314#issuecomment-795642646)?

Bodigrim:
I understand the temptation, but I'd prefer to drop GHC < 8.0 in a dedicated PR.

You can throw `-fno-warn-redundant-imports`, so that you can add missing imports without CPP.

Bodigrim (approved):

PR #325, opened 2021-04-02
Use unsafeWithForeignPtr when available

bgamari:
It turns out that all uses of `withForeignPtr` can safely use
`unsafeWithForeignPtr`.

I believe that this should resolve the regression noted in
[GHC #19557](https://gitlab.haskell.org/ghc/ghc/-/issues/19557)
although I am still benchmarking


Bodigrim:
@bgamari could you please fix failed builds and rebase?

Bodigrim (approved):
This does not seem to fix a performance regression in GHC 9.0, but is useful nevertheless. Thanks!

Boarders:
Thanks @bgamari!

Lysxia (approved):

PR #326, opened 2021-04-02
Add windows, macOS and freebsd jobs

ethercrow:


.github/workflows/windows_and_macOS.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: cabal update
+    - uses: actions/cache@v2.1.4
+      name: Cache cabal stuff
+      with:
+        path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}
+    - name: Test
+      run: cabal test
+    - name: Bench
+      run: cabal bench
```
phadej, line 35: Is it good idea to **run** benchmarks on CI? The `text` benchmark suite isn't very fast to run (and need benchmark dta checkout too, IIRC)

Bodigrim: I think `cabal bench --benchmark-option=-l` would be good enough, just to check that benchmarks are still buildable under Windows/MacOS.

ethercrow: I started by copypasting bytestring's CI config, I agree that running `cabal bench` for text is too much.

ethercrow:
Curiously, the freebsd job complained about this:

```
          t_toTitle_1stNotLower:   FAIL
            *** Failed! Falsified (after 55 tests and 5 shrinks):
            "\4322"
            Use --quickcheck-replay=174726 to reproduce.
```

Bodigrim:
Yeah, I've seen `t_toTitle_1stNotLower` test failure couple of times before :(

ethercrow:
There's seems to be a windows specific issue:

```
hGetContents_crash:            FAIL
      Exception: .\cra4F80.txt: DeleteFile "\\\\?\\D:\\a\\text\\text\\cra4F80.txt": permission denied (The process cannot access the file because it is being used by another process.)
```

@Bodigrim didn't bytestring have something similar?

Bodigrim:
@ethercrow sounds similar to https://github.com/haskell/bytestring/pull/320

Bodigrim (approved):
Looks good to me, thanks!

ethercrow:
Now we know that the benchmark suite doesn't build on Windows:

```
benchmarks\cbits-bench\time_iconv.c:1:10: error:
     fatal error: iconv.h: No such file or directory
        1 | #include <iconv.h>
          |          ^~~~~~~~~
  |
1 | #include <iconv.h>

  |          ^
compilation terminated.
`gcc.exe' failed in phase `C Compiler'. (Exit code: 1)
```

Bodigrim:
Ahh, interesting. I think it's acceptable to remove `iconv` benchmark with all relevant `cbits`: it is not particularly important nowadays. We should not expect `iconv` being present out of the box on non-Windows platforms either, and I do not feel benchmarking it is worth environmental complications.

phadej:
FWIW, `iconv` is always there on non-Windows platforms. GHC relies on it. https://hackage.haskell.org/package/base-4.15.0.0/docs/src/GHC-IO-Encoding-Iconv.html#iconvEncoding is one example.

In other words, you can reasonably expect `iconv` to be present on non-Windows platforms.

ethercrow:
For now I just removed building benchmarks on windows.

Bodigrim (approved):

Lysxia (approved):

PR #333, opened 2021-05-02
Fix UTF-8 decoding of lazy bytestrings

### LINKED ISSUE #330: Inconsistent UTF8 decoding behaviour based on the underlying chunking
Depending on the way the input is chunked, `decodeUtf8With` returns different results for the "same" bytestring. See:

```haskell
> import qualified Data.Text.Lazy.Encoding as T
> import qualified Data.Text.Encoding.Error as T
> import qualified Data.ByteString as BS
> import qualified Data.ByteString.Lazy as BL
> let bs1 = BL.fromChunks [BS.pack [194], BS.pack [97, 98, 99]]
> let bs2 = BL.fromChunks [BS.pack [194, 97, 98, 99]]
> bs1
"\194abc"
> bs2
"\194abc"
> bs1 == bs2
True
> T.decodeUtf8With T.lenientDecode bs1
"\65533bc"
> T.decodeUtf8With T.lenientDecode bs2
"\65533abc"
```

Another example:

```haskell
> let Some y l r = streamDecodeUtf8With lenientDecode "\194"
> (y, l)
("","\194")
> let Some y' l' r' = r "abcde"
> (y', l')
("\65533bcde","")
```

but:

```haskell
> let Some y l r = streamDecodeUtf8With lenientDecode "\194abcde"
> (y, l)
("\65533abcde","")
```

I noticed this property testing a function using `streamDecodeUtf8With` against a simpler one using `decodeUtf8With` on strict `Text`; but the same issue appeared on other places too.

To me, this sounds like a bug; since it breaks equational reasoning; but if it is the expected behaviour for some reason it should be documented prominently around `decodeUtf8` functions.

This issue looks vaguely relevant: https://github.com/haskell/text/issues/60

Lysxia:
Fixes #330.

I found another issue that's not yet fixed: both strict and lazy `decodeUtf8With` are actually memory unsafe if you use a bad `onErr` argument. We allocate a destination buffer with 2x the number of bytes from the original bytestrings, but by using an `onErr` function which replaces any invalid byte with a `Char` which is a surrogate pair in UTF-16, it possible to blow up the size taken by a `Text` to 4x. In practice, `onErr` is almost always `lenientDecode` though, so perhaps a better solution than allocating more memory is to either hide `decodeUtf8With` or clamp the range of `onErr`.

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -64,6 +64,7 @@ import Control.Monad.ST (runST)
 import Data.Bits ((.&.))
 import Data.ByteString as B
 import qualified Data.ByteString.Internal as B
+import Data.Foldable (traverse_)
 import Data.Text.Encoding.Error (OnDecodeError, UnicodeException, strictDecode)
 import Data.Text.Internal (Text(..), safe, text)
 import Data.Text.Internal.Functions
@@ -275,7 +276,7 @@ newtype DecoderState = DecoderState Word32 deriving (Eq, Show, Num, Storable)
 streamDecodeUtf8 :: ByteString -> Decoding
 streamDecodeUtf8 = streamDecodeUtf8With strictDecode
 
--- | Decode, in a stream oriented way, a 'ByteString' containing UTF-8
+-- | Decode, in a stream oriented way, a lazy 'ByteString' containing UTF-8
 -- encoded text.
 --
 -- @since 1.0.0.0
@@ -283,11 +284,14 @@ streamDecodeUtf8With :: OnDecodeError -> ByteString -> Decoding
 streamDecodeUtf8With onErr = decodeChunk B.empty 0 0
  where
   -- We create a slightly larger than necessary buffer to accommodate a
-  -- potential surrogate pair started in the last buffer
+  -- potential surrogate pair started in the last buffer (@undecoded0@), or
+  -- replacement characters for each byte in @undecoded0@ if the
+  -- sequence turns out to be invalid. There can be up to three bytes there,
+  -- hence we allocate @len+3@ 16-bit words.
   decodeChunk :: ByteString -> CodePoint -> DecoderState -> ByteString
               -> Decoding
   decodeChunk undecoded0 codepoint0 state0 bs = withBS bs aux where
-    aux fp len = runST $ (unsafeIOToST . decodeChunkToBuffer) =<< A.new (len+1)
+    aux fp len = runST $ (unsafeIOToST . decodeChunkToBuffer) =<< A.new (len+3)
        where
         decodeChunkToBuffer :: A.MArray s -> IO Decoding
         decodeChunkToBuffer dest = unsafeWithForeignPtr fp $ \ptr ->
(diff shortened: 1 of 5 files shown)


Lysxia:
I don't understand the doctest errors on GHC 8+, and on GHC 7+ the errors are that old versions of bytestring did not have `toStrict`, which I guess I could fix by concatenating `toChunks` instead.

src/Data/Text/Encoding.hs
```diff
@@ -285,19 +285,22 @@ newtype DecoderState = DecoderState Word32 deriving (Eq, Show, Num, Storable)
...
 -- encoded text.
 --
 -- @since 1.0.0.0
 streamDecodeUtf8With :: OnDecodeError -> ByteString -> Decoding
 streamDecodeUtf8With onErr = decodeChunk B.empty 0 0
  where
   -- We create a slightly larger than necessary buffer to accommodate a
-  -- potential surrogate pair started in the last buffer
+  -- potential surrogate pair started in the last buffer (@undecode0@), or
+  -- replacement characters for each byte in @undecode0@ if the
```
Bodigrim, line 297: Nitpick: it's `undecoded0`, not `undecode0`. Same at the previous line.

src/Data/Text/Encoding.hs
```diff
@@ -307,23 +310,31 @@ streamDecodeUtf8With onErr = decodeChunk B.empty 0 0
...
-                          destOff <- peek destOffPtr
-                          w <- unsafeSTToIO $
-                               unsafeWrite dest (fromIntegral destOff) (safe c)
-                          poke destOffPtr (destOff + fromIntegral w)
-                          loop $ curPtr' `plusPtr` 1
+                      if ptr == lastPtr && prevState /= UTF8_ACCEPT then do
+                        -- If we can't complete the sequence undecode0 from the previous chunk,
+                        -- we invalidate the bytes from undecode0 and retry decoding
+                        -- the current chunk from the initial state.
+                        forBS undecoded0 skipByte
```
Bodigrim, line 333: I'd just `traverse skipByte $ B.unpack undecoded0`, just to avoid an indirection to `forBS` definition. I don't think there are serious performance implications.

Bodigrim (commented):
Doctests are failing in `master` as well (I guess it has something to do with a new version of `doctest` package?.. Dunno)

Could you please check a coverage report to ensure that all lines are well-tested?

Bodigrim:
That's probably fine, I guess. 

Looks good to me except GHC < 7.6 builds.

tests/Tests/QuickCheckUtils.hs
```diff
@@ -79,6 +80,10 @@ instance Arbitrary B.ByteString where
     arbitrary     = B.pack `fmap` arbitrary
     shrink        = map B.pack . shrink . B.unpack
 
+instance Arbitrary BL.ByteString where
+    arbitrary = BL.fromChunks <$> arbitrary
```
Bodigrim, line 84:
```suggestion
    arbitrary = oneof
      [ BL.fromChunks <$> arbitrary
      -- so that a single utf8 code point could appear split over up to 4 chunks
      , BL.fromChunks . map B.singleton <$> arbitrary
      -- so that a code point with 4 byte long utf8 representation
      -- could appear split over 3 non-singleton chunks
      , (\a b c -> BL.fromChunks [a, b, c])
        <$> arbitrary
        <*> ((\a b -> B.pack [a, b]) <$> arbitrary <*> arbitrary)
        <*> arbitrary
      ]
```
So that we test even most malicious strings.

Lysxia: Good idea. I also tweaked the generators to not use `arbitrary` for `Word8`, which is size-based (which makes short strings ASCII only), opting instead for a simple uniform distribution.

Bodigrim (approved):
Nice!

Bodigrim (dismissed):

Lysxia:
(Found a silly space.)

Bodigrim (approved):

Bodigrim:
@Lysxia could you please resolve a conflict?

Bodigrim (approved):

PR #336, opened 2021-05-08
Drop support of GHC < 8

Bodigrim:
As discussed at previous maintainers meetings. 
A utf8 journey would go smoother without worrying about ancient GHCs.

CC @chessai @emilypi

### DIFF
--- a/src/Data/Text/Internal.hs
+++ b/src/Data/Text/Internal.hs
@@ -20,10 +20,6 @@
 -- with the internals, as the functions here do just about nothing to
 -- preserve data invariants.  You have been warned!
 
-#if defined(__GLASGOW_HASKELL__) && !defined(__HADDOCK__)
-#include "MachDeps.h"
-#endif
-
 module Data.Text.Internal
     (
     -- * Types
@@ -124,11 +120,11 @@ firstf _  Nothing      = Nothing
 -- | Checked multiplication.  Calls 'error' if the result would
 -- overflow.
 mul :: Int -> Int -> Int
-#if WORD_SIZE_IN_BITS == 64
-mul a b = fromIntegral $ fromIntegral a `mul64` fromIntegral b
-#else
-mul a b = fromIntegral $ fromIntegral a `mul32` fromIntegral b
-#endif
+mul a b
+  | finiteBitSize (0 :: Word) == 64
+  = fromIntegral $ fromIntegral a `mul64` fromIntegral b
+  | otherwise
+  = fromIntegral $ fromIntegral a `mul32` fromIntegral b
 {-# INLINE mul #-}
 infixl 7 `mul`
 
--- a/src/Data/Text/Internal/Unsafe/Shift.hs
+++ b/src/Data/Text/Internal/Unsafe/Shift.hs
@@ -1,8 +1,5 @@
 {-# LANGUAGE CPP #-}
-#if MIN_VERSION_base(4,5,0)
--- base-4.5.0 is 7.4, default sigs introduced in 7.2
 {-# LANGUAGE DefaultSignatures #-}
-#endif
 {-# LANGUAGE MagicHash #-}
 
 -- |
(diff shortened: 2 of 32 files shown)


chessai (approved):

src/Data/Text/Internal.hs
```diff
@@ -124,11 +120,11 @@ firstf _  Nothing      = Nothing
...
-#if WORD_SIZE_IN_BITS == 64
-mul a b = fromIntegral $ fromIntegral a `mul64` fromIntegral b
-#else
-mul a b = fromIntegral $ fromIntegral a `mul32` fromIntegral b
-#endif
+mul a b
+  | finiteBitSize (0 :: Word) == 64
+  = fromIntegral $ fromIntegral a `mul64` fromIntegral b
+  | otherwise
+  = fromIntegral $ fromIntegral a `mul32` fromIntegral b
```
emilypi, line 127: This introduces some new code. Is there something wrong with keeping the machdeps import and CPP'ing these definitions?

emilypi (approved):
Looks like a great start. Just one non-blocking question.

src/Data/Text/Internal.hs
phadej: GHC constant folds `finiteBitSize (0 :: Word)` branches.

text.cabal
```diff
@@ -146,35 +131,18 @@ library
...
-  else
-    build-depends: bytestring         >= 0.10.4 && < 0.12
-
   ghc-options: -Wall -fwarn-tabs -funbox-strict-fields -O2
   if flag(developer)
     ghc-options: -fno-ignore-asserts
     cpp-options: -DASSERTS
 
-  if impl(ghc >= 8.11)
-    build-depends: ghc-bignum
```
phadej, line 167: removal of this is very nice.

src/Data/Text/Internal.hs
Bodigrim: There is very little public infrastructure available to test 32 bit platforms, e. g., AFAIK no CI service has such an option. Using `finiteBitSize` allows at least to check that 32 bit code path is well typed, while with CPP it could even do not parse.

chessai: To my knowledge, a nix one-liner is the easiest way to emulate 32-bit platforms on 64-bit ones. Simply specify the "i686-linux" system in your nixpkgs. Here is a small example illustrating this:

```
$ cat shell.nix
with (import <nixpkgs> { config = {}; overlays = []; system = "i686-linux"; });

mkShell {
  name = "32-bit-nix-emulation-example";

  GHC_INCLUDE_PATH = "${pkgs.haskell.compiler.ghc8104}/lib/ghc-8.10.4/include";

  buildInputs = [
    gcc
  ];
}

$ cat main.c
#include <stdio.h>
#include "MachDeps.h"

int main() {
  printf("%i", WORD_SIZE_IN_BITS);
}

$ nix-shell

nix-shell$ gcc -Wall -I $GHC_INCLUDE_PATH main.c -o main && ./main
32

nix-shell$ exit

$ #edit shell.nix to say `system = x86_64-linux` now

$ gcc -Wall -I $GHC_INCLUDE_PATH main.c -o main && ./main
64
```

TLDR it's sufficient to specify just "system = i686-linux" to emulate a 32-bit platform. I have used this in the past to test things like chronos (the time library) on 32-bit.

Lysxia:
Do we want to merge that soon or should we rather wait so the next release (which has quite some stuff already) remains more widely compatible?

Bodigrim:
Feel free to release current `master`, if you feel motivated enough. I personally would not bother. Otherwise I need this PR merged early to start working on utf8 transition.

src/Data/Text/Internal/Unsafe/Shift.hs
```diff
@@ -1,8 +1,5 @@
 {-# LANGUAGE CPP #-}
```
Lysxia, line 1: Can't we get rid of this whole file? It says it's a workaround for GHC 6... I'd be fine punting this as a later TODO though.

Lysxia (approved):

src/Data/Text/Internal/Unsafe/Shift.hs
emilypi: I'd say let's move the bigger code-change work to a follow up to unblock @Bodigrim

Issue #341, opened 2021-05-13
Potential decoding helpers?


pdedmon:
Would any of the following definitions be worth adding to Data.Text.Encoding?

```
decodeUtf8Either :: ByteString -> Either UnicodeException Text
decodeUtf8Either = decodeUtf8'
```
```
decodeUtf8Maybe :: ByteString -> Maybe Text
decodeUtf8Maybe = either (const Nothing) Just . decodeUtf8'
```

```
decodeUtf8Lenient :: ByteString -> Text
decodeUtf8Lenient = decodeUtf8With lenientDecode
```

These are simple enough to define ad hoc, but I thought I'd ask anyway, to save myself from having to do so :slightly_smiling_face:.

phadej:
Virtually all my project specific preludes introduce `decodeUtf8Lenient`, so :+1: from me.

pdedmon:
Cool, I'll get a pull request together.

Issue #247, opened 2019-01-20
`decodeUtf8` / `decodeUtf8'` situation


seagreen:
There are a few issues with the current situation:

+ Decoding/encoding from UTF-8 is a really common use case, but requires a separate import (`Data.Text.Encoding`)
+ The best name (`decodeUtf8`) is taken by the partial version of the function
+ `decodeUtf8'` doesn't mean anything, you just have to know the `'` means it's the total version

Is there anything we could do to improve this situation? It doesn't bother me, but it's tough on people that are new to haskell. I know we're constrained by backwards compatibility here because so many packages use `text`, but I thought it would be worth thinking about.

hvr:
The one thing we can do with little risk is improve the documentation...

seagreen:
Opened an issue for docs: https://github.com/haskell/text/issues/248

seagreen:
That can only do so much when the best names are taken up by partial variants of course. Is there interest in adding new functions with names like `decodeUtf8Partial`? Then (on a very long timescale) we could add deprecation notices to the `decodeUtf8` style partial functions and get people to either switch to the `*Partial` variants or handle failure.

This would definitely improve this library. It might improve the ecosystem in general as well (since I bet a lot of people are using decodeUtf8 without realizing it can throw exceptions in pure code).

hvr:
The problem I see is with the "very long timescale" -- what if the timescale is 10 years, does it still make sense to do this? as you point out `text` has way too many consumers which might be affected; way too much for a change which is primarily motivated by aesthetics/discoverability (which I do sympathize with though). This isn't the only instance where there's regret in a naming or type-signature choice that was done a decade ago, and where we'd just have to stick with it until there's a technically motivated excuse to fix it up presents itself...

Btw, the same argument about these functions being stuffed away in a separate module could equally be made about the I/O ops being hidden away in `Data.Text.IO`, couldn't it?

cblp:
Partial functions are traditionally called "unsafe..." in Haskell. Please rename decodeUtf8 to unsafeDecodeUtf8.

michaelpj:
I'd like to re-raise this issue. While it would be nice to rename `decodeUtf8`, I'd want to suggest a much more direct and uncomfortable approach:

1. Deprecate `decodeUtf8`.
2. Provide `unsafeDecodeUtf8` with the behaviour of the current `decodeUtf8`.
 
[...]

PR #342, opened 2021-05-15
add some utf8 decoding convenience functions

pdedmon:
Addresses #341.

Let me know if I should add any INLINE pragmas or "since" notes in the docs.

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -35,6 +35,7 @@ module Data.Text.Encoding
 
     -- ** Controllable error handling
     , decodeUtf8With
+    , decodeUtf8Lenient
     , decodeUtf16LEWith
     , decodeUtf16BEWith
     , decodeUtf32LEWith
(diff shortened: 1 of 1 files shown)


Lysxia (dismissed):

Bodigrim:
`decodeUtf8Lenient` is fine, but I'm really unsure how much our vocabulary benefit from a synonym and a trivial composition. I do not object however.

Lysxia:
It's arguable that `decodeUtf8'` is not a good name ([another issue](https://github.com/haskell/text/issues/247) complaining about it), whereas if you see `decodeUtf8Either` you can better guess what it does. I do share your reservations about also having a `Maybe` version though...

pdedmon:
Shall I reduce this down to just `decodeUtf8Lenient` then? I don't feel too strongly about the others, though I do agree with the sentiment around `decodeUtf8'`.

Lysxia:
Let's do that, yes.

pdedmon:
Done and rebased.

Lysxia (approved):

PR #343, opened 2021-05-23
Preparatory work to facilitate utf8 transition

Bodigrim:
The main work on utf8 transition is likely to be a pretty long-lived branch, but there are several preparatory changes, which improve development experience in general, and I'd like to merge them early, so that I do not spend time rebasing them. 

When merging, please **rebase, do not squash**. Each commit is a separate, orthogonal change here, but raising 6 PRs seems an overkill.


Bodigrim:
Discussion points:

1. One might argue that it's preferable to collect all integer conversions in a single module instead of scattering them around. I think there is not much overlap between modules at the moment to warrant such move, but I would not mind to do so, if others feel it as a better alternative.

2. Instead of repeating `#ifdef ASSERT HasCallStack => #endif Bla -> Bla -> Bla` everywhere, we can define in a separate module `type HasCallStack' = #ifdef ASSERT HasCallStack #else () #endif` and then just use `HasCallStack' => Bla -> Bla -> Bla`. Does it sound appealing to do so?

Lysxia (approved):
LGTM

Lysxia:
Re: discussion points. I think the way you did things is fine. In particular for 2, refactoring into a type synonym unfortunately makes it visible to haddock.

PR #345, opened 2021-05-28
Refactor tests

Bodigrim:
This is mostly to replace `forAll` (without `shrink`) with a newtype, supporting shrinking. Shrinking is vital for actionable property testing. I also put an effort into making `t_utf8_err` more friendly, with detailed reporting of counterexamples. I hit some issues in utf8 branch, and it was quite difficult to debug otherwise.

### DIFF
--- a/tests/Tests/Properties/Text.hs
+++ b/tests/Tests/Properties/Text.hs
@@ -162,15 +162,16 @@ tl_find p         = L.find p      `eqP` TL.find p
 t_partition p     = L.partition p `eqP` (unpack2 . T.partition p)
 tl_partition p    = L.partition p `eqP` (unpack2 . TL.partition p)
 
-sf_index p s      = forAll (choose (-l,l*2))
-                    ((L.filter p s L.!!) `eq` S.index (S.filter p $ packS s))
+sf_index p s i    = ((L.filter p s L.!!) `eq` S.index (S.filter p $ packS s)) j
     where l = L.length s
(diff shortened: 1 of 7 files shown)


tests/Tests/Properties/Text.hs
```diff
@@ -169,15 +162,16 @@ tl_find p         = L.find p      `eqP` TL.find p
 t_partition p     = L.partition p `eqP` (unpack2 . T.partition p)
 tl_partition p    = L.partition p `eqP` (unpack2 . TL.partition p)
 
-sf_index p s      = forAll (choose (-l,l*2))
-                    ((L.filter p s L.!!) `eq` S.index (S.filter p $ packS s))
+sf_index p s i    = ((L.filter p s L.!!) `eq` S.index (S.filter p $ packS s)) j
     where l = L.length s
-t_index s         = forAll (choose (-l,l*2)) ((s L.!!) `eq` T.index (packS s))
+          j = if l == 0 then 0 else i `mod` (3 * l) - l
+t_index s i       = ((s L.!!) `eq` T.index (packS s)) j
```
Lysxia, line 168: This is OK (I was wondering whether `i` might be too small because it's no longer explicitly using `choose`, but both `s` and `i` are going to be generated with the same size parameter, so by their `Arbitrary` instances, `s`'s length is going to be `choose (0, n)` and `i` is going to be `choose (-n, n)`).

Lysxia (approved):

PR #346, opened 2021-06-09
Tweak benchmark suite

Bodigrim:
While working on UTF8 transition, I run and rerun benchmarks many times. It appears that some of them measure performance of `String` and `ByteString`, which does not bear much relevance for `text` package. This PR removes such benchmarks, because we are just wasting time and energy.

I also changed hierarchy of benchmarks to make proper use of `bgroup`: previously group names were appended to individual benchmark names as suffixes, e. g., `concat+japanese` instead of proper `japanese.concat`.


benchmarks/haskell/Benchmarks/Search.hs
```diff
@@ -12,42 +12,30 @@ module Benchmarks.Search
...
+benchmark needleT ~(t, tl) =
     bgroup "FileIndices"
-        [ bench "ByteString"     $ whnf (byteString needleB)     b
-        , bench "LazyByteString" $ whnf (lazyByteString needleB) bl
-        , bench "Text"           $ whnf (text needleT)           t
+        [ bench "Text"           $ whnf (text needleT)           t
         , bench "LazyText"       $ whnf (lazyText needleTL)      tl
         ]
   where
     needleB = T.encodeUtf8 needleT
```
Lysxia, line 36: Seems like this can be cleaned up.

Lysxia (approved):

PR #352, opened 2021-06-21
CI tweaks

Bodigrim:
1. Regenerate CI config using newer `haskell-ci`.
2. Do not restrict CI builds to `master` branch only.
3. I'm tired of dependabot notifying us about every minor update of any component. This churn is not worth it, I think.
4. Add a build with `developer` flag enabled.


Bodigrim:
This is all green now.

Lysxia (approved):

PR #354, opened 2021-07-01
Assorted improvements

Bodigrim:
`Data.Text.Internal.Functions` in `text-1.2.4.1` contains only one function `intersperse`, whose definition actually matches `Data.List` since `base-4.4`. It makes no sense to carry on with this module. 

`unsafeWithForeignPtr` was added to the same module in #325, but this change has not been released yet, and I think `Data.Text.Internal.Unsafe` is a better place for such function. 

I discovered that our tests intentionally disable rewrite rules. This is extremely weird, because it means that rewrite rules (which are always enabled for users) remain completely untested. I don't know good reasons to test configuration with disabled rules, so I enabled them unconditionally. Potentially we can add a dedicated CI build with `cabal test --ghc-option=-fno-enable-rewrite-rules`, but I'm not sure it's worth it.

`words` were unlikely to be thoroughly tested before; I've added a special QuickCheck generator, which creates strings with many interestring space characters. I also written some additional tests for other functions. 

I was looking for Apple Silicon CI, but it seems there are no free services of this kind so far. So I added drone.io for Ubuntu ARM build, which is better than nothing. E. g., https://cloud.drone.io/Bodigrim/text/4/1/2


tests/Tests/Properties/Text.hs
```diff
@@ -38,18 +38,27 @@ tl_unstreamChunks x = f 11 x === f 1000 x
...
 s_intercalate c   = (L.intercalate c . unSqrt) `eq`
                     (unpackS . S.intercalate (packS c) . map packS . unSqrt)
 t_intercalate c   = (L.intercalate c . unSqrt) `eq`
                     (unpackS . T.intercalate (packS c) . map packS . unSqrt)
 tl_intercalate c  = (L.intercalate c . unSqrt) `eq`
                     (unpackS . TL.intercalate (TL.pack c) . map TL.pack . unSqrt)
+t_length_intercalate c  = (L.length . L.intercalate c . unSqrt) `eq`
+                    (T.length . T.intercalate (packS c) . map packS . unSqrt)
+tl_length_intercalate c = (L.genericLength . L.intercalate c . unSqrt) `eq`
+                    (TL.length . TL.intercalate (TL.pack c) . map TL.pack . unSqrt)
```
Lysxia, line 61: Is this not redundant with the tests without length? There may be some reason but I don't quite see it.

Bodigrim: Treat them as tests for `length`, not for `intercalate`. They are not immediately useful with current implementation of `length`, but I found them useful for other purposes, so it is easier to submit them separately.

symbiont-callan-mcgill:
Thanks, as always, for the many excellent improvements @Bodigrim !

Boarders:
Thanks, as always, for the many improvements @Bodigrim!

Lysxia (approved):

PR #355, opened 2021-07-14
Yet another omnibus

Bodigrim:
This is mostly focused on improving and cleaning up the test suite. 

I was interested to remove `vector` dependency from benchmarks, because I build `text` against GHC HEAD, and rebuilding `vector` over and over again is not worth it.


Boarders:
Don't quite follow all the details here, but overall looks like a good direction!

Boarders (approved):

parsonsmatt (approved):
Everything looks like an improvement to me 👍🏻

PR #366, opened 2021-09-08
Use copyToPointer in encodeUtf8Builder

alexbiehl:
Instead of copying each character separately by using `encodeUtf8BuilderEscaped` we instead resort to memcpy to copy the bytes into the target buffer more efficiently.

This PR is based off similar functions in the ByteString package: 

https://github.com/haskell/bytestring/blob/ba3046d9f5bf5cac1246c39f519b8be7ac11f49c/Data/ByteString/Builder/Internal.hs#L832-L845

https://github.com/haskell/bytestring/blob/ba3046d9f5bf5cac1246c39f519b8be7ac11f49c/Data/ByteString/Builder/Internal.hs#L779-L796

### DIFF
--- a/benchmarks/haskell/Benchmarks/EncodeUtf8.hs
+++ b/benchmarks/haskell/Benchmarks/EncodeUtf8.hs
@@ -10,8 +10,10 @@ module Benchmarks.EncodeUtf8
     ( benchmark
     ) where
 
-import Test.Tasty.Bench (Benchmark, bgroup, bench, whnf)
+import Test.Tasty.Bench (Benchmark, bgroup, bench, nf, whnf)
 import qualified Data.ByteString as B
+import qualified Data.ByteString.Builder as B
+import qualified Data.ByteString.Builder.Prim as BP
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
@@ -23,6 +25,8 @@ benchmark name string =
     bgroup name
         [ bench "Text"     $ whnf (B.length . T.encodeUtf8)   text
         , bench "LazyText" $ whnf (BL.length . TL.encodeUtf8) lazyText
+        , bench "Text/encodeUtf8Builder" $ nf (B.toLazyByteString . T.encodeUtf8Builder) text
+        , bench "Text/encodeUtf8BuilderEscaped" $ nf (B.toLazyByteString . T.encodeUtf8BuilderEscaped (BP.liftFixedToBounded BP.word8)) text
         ]
   where
     -- The string in different formats
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -391,7 +391,37 @@ decodeUtf8Lenient = decodeUtf8With lenientDecode
 --
 -- @since 1.1.0.0
 encodeUtf8Builder :: Text -> B.Builder
-encodeUtf8Builder = encodeUtf8BuilderEscaped (BP.liftFixedToBounded BP.word8)
+encodeUtf8Builder =
+    -- manual eta-expansion to ensure inlining works as expected
+    \txt -> B.builder (step txt)
+  where
+    step txt@(Text arr off len) !k br@(B.BufferRange op ope)
+      -- Ensure that the common case is not recursive and therefore yields
+      -- better code.
+      | op' <= ope = do
+          unsafeSTToIO $ A.copyToPointer arr off op len
+          k (B.BufferRange op' ope)
+      | otherwise = textCopyStep txt k br
+      where
+        op' = op `plusPtr` len
+{-# INLINE encodeUtf8Builder #-}
+
+textCopyStep :: Text -> B.BuildStep a -> B.BuildStep a
+textCopyStep (Text arr off len) k =
+    go off (off + len)
+  where
+    go !ip !ipe (B.BufferRange op ope)
+      | inpRemaining <= outRemaining = do
+          unsafeSTToIO $ A.copyToPointer arr ip op inpRemaining
+          let !br = B.BufferRange (op `plusPtr` inpRemaining) ope
+          k br
+      | otherwise = do
+          unsafeSTToIO $ A.copyToPointer arr ip op outRemaining
+          let !ip' = ip + outRemaining
+          return $ B.bufferFull 1 ope (go ip' ipe)
+      where
+        outRemaining = ope `minusPtr` op
+        inpRemaining = ipe - ip
 
 -- | Encode text using UTF-8 encoding and escape the ASCII characters using
 -- a 'BP.BoundedPrim'.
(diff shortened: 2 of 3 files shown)


Boarders:
Did you benchmark the change by any chance?

alexbiehl:
@Boarders No, I did not. But shouldn't be hard to set something up I guess?

Boarders:
@alexbiehl I'm sure it is better but always good to confirm.

Bodigrim:
It appears we do not have any tests for `encodeUtf8Builder`. @alexbiehl could you possibly add some?

alexbiehl:
I set up a small benchmark for encodeUtf8Builder: 

```
All
  EncodeUtf8
    non-ASCII
      Text:                          OK (2.19s)
        251  μs ±  23 μs
      LazyText:                      OK (3.14s)
        190  μs ±  16 μs
      Text/encodeUtf8Builder:        OK (20.15s)
        298  ns ±  13 ns
      Text/encodeUtf8BuilderEscaped: OK (2.07s)
        3.91 μs ± 213 ns
    ASCII
      Text:                          OK (1.34s)
        78.2 μs ± 7.2 μs
      LazyText:                      OK (8.72s)
        65.7 μs ± 1.8 μs
      Text/encodeUtf8Builder:        OK (1.30s)
        293  ns ±  27 ns
      Text/encodeUtf8BuilderEscaped: OK (2.28s)
        8.67 μs ± 462 ns

All 8 tests passed (41.26s)
```

benchmarks/haskell/Benchmarks/EncodeUtf8.hs
```diff
@@ -23,6 +25,8 @@ benchmark name string =
     bgroup name
         [ bench "Text"     $ whnf (B.length . T.encodeUtf8)   text
         , bench "LazyText" $ whnf (BL.length . TL.encodeUtf8) lazyText
+        , bench "Text/encodeUtf8Builder" $ whnf (B.toLazyByteString . T.encodeUtf8Builder) text
+        , bench "Text/encodeUtf8BuilderEscaped" $ whnf (B.toLazyByteString . T.encodeUtf8BuilderEscaped (BP.liftFixedToBounded BP.word8)) text
```
Bodigrim, line 29: This looks a bit weird: `whnf` of a lazy bytestring does not trigger anything except the first chunk.

tests/Tests/Properties/Transcoding.hs
```diff
@@ -49,6 +52,11 @@ tl_utf32LE   = (EL.decodeUtf32LE . EL.encodeUtf32LE) `eq` id
 t_utf32BE    = (E.decodeUtf32BE . E.encodeUtf32BE) `eq` id
 tl_utf32BE   = (EL.decodeUtf32BE . EL.encodeUtf32BE) `eq` id
 
+t_encodeUtf8Builder = (E.decodeUtf8 . BL.toStrict . B.toLazyByteString . E.encodeUtf8Builder) `eq` id
+t_encodeUtf8Builder_sanity t =
+  (B.toLazyByteString . E.encodeUtf8Builder) t ==
```
Bodigrim, line 57:
```suggestion
  (B.toLazyByteString . E.encodeUtf8Builder) t ===
```

benchmarks/haskell/Benchmarks/EncodeUtf8.hs
alexbiehl: Ah, right!

alexbiehl:
Much more realistic now: 

```
All
  EncodeUtf8
    non-ASCII
      Text:                          OK (2.24s)
        262  μs ±  15 μs
      LazyText:                      OK (1.78s)
        208  μs ±  19 μs
      Text/encodeUtf8Builder:        OK (9.07s)
        272  μs ± 6.2 μs
      Text/encodeUtf8BuilderEscaped: OK (1.53s)
        2.94 ms ± 247 μs
    ASCII
      Text:                          OK (1.41s)
        81.7 μs ± 6.8 μs
      LazyText:                      OK (4.64s)
        69.3 μs ± 3.7 μs
      Text/encodeUtf8Builder:        OK (2.89s)
        86.5 μs ± 6.3 μs
      Text/encodeUtf8BuilderEscaped: OK (2.41s)
        2.33 ms ± 154 μs

```

Boarders:
@alexbiehl Sorry to be a pain but could you run the benchmark on the previous implementation just to see how this changes things?

alexbiehl:
@Boarders You can tell by looking at the `Text/encodeUtf8BuilderEscaped` line.

Boarders:
Looks good, thank you very much for adding it!

src/Data/Text/Encoding.hs
```diff
@@ -391,7 +391,37 @@ decodeUtf8Lenient = decodeUtf8With lenientDecode
...
+          unsafeSTToIO $ A.copyToPointer arr off op inpRemaining
+          let !br = B.BufferRange (op `plusPtr` inpRemaining) ope
+          k br
+      | otherwise = do
+          unsafeSTToIO $ A.copyToPointer arr off op outRemaining
+          let !off' = off + outRemaining
+          return $ B.bufferFull 1 ope (go off')
+      where
+        outRemaining = ope `minusPtr` op
+        inpRemaining = len - off
```
Bodigrim, line 424: @alexbiehl could you please add a test, which exercises `encodeUtf8Builder` with non-zero `off`?

alexbiehl: Is there any existing test that does so? No idea how to test that effectively.

alexbiehl: Ah, got it.

alexbiehl: And indeed it catched the bug!

src/Data/Text/Encoding.hs
```diff
@@ -391,7 +391,37 @@ decodeUtf8Lenient = decodeUtf8With lenientDecode
...
+  where
+    go !ip !ipe (B.BufferRange op ope)
+      | inpRemaining <= outRemaining = do
+          unsafeSTToIO $ A.copyToPointer arr ip op inpRemaining
+          let !br = B.BufferRange (op `plusPtr` inpRemaining) ope
+          k br
+      | otherwise = do
+          unsafeSTToIO $ A.copyToPointer arr ip op outRemaining
+          let !ip' = ip + outRemaining
+          return $ B.bufferFull 1 ope (go ip' ipe)
```
Bodigrim, line 421: It could be that I terribly misunderstand how `bufferFull` works, but why do we pass `1` instead of `inpRemaining - outRemaining`? We could have avoided any recursive functions at all, no?

alexbiehl: I think it signals „I need at least one byte to make progress“. As stated in the description this is copied more or less from the ByteString and ShortByteString cases.

alexbiehl: Comment on bufferFull arg says 
„Minimal size of next 'BufferRange'.“

Bodigrim: Yeah, I have the same question about implementations in `bytestring` :)

alexbiehl: So to me it looks like conservative default. If you’d pass the the length of the remaining string you’d put assumptions on the upper bound of the size of the buffer you are writing to while we really want the driver to be in charge of the buffer alone. I could def. see a compromise of passing upper and lower bound together so that the driver can decide how much to allocate more precisely.

Bodigrim (approved):

Lysxia (approved):

PR #367, opened 2021-09-11
Haskell-CI for 9.2-rc1

andreasabel:
Test also GHC 9.2-rc1 under ubuntu.

Generated automatically with `haskell-ci` from `tested-with: GHC-9.2.0.20210821`.

Successful run at https://github.com/andreasabel/text/runs/3576075282?check_suite_focus=true

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.13.20210606
+# version: 0.13.20210901
 #
-# REGENDATA ("0.13.20210606",["github","--config=cabal.haskell-ci","text.cabal"])
+# REGENDATA ("0.13.20210901",["github","--config=cabal.haskell-ci","text.cabal"])
 #
 name: Haskell-CI
 on:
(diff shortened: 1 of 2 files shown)


Bodigrim:
I'm ambivalent here: it would be nice to test against GHC 9.2 RC, but it means we'd have to update CI config once again when GHC 9.2.1 is out. I'm happy for other maintainers to decide.

andreasabel:
@Bodigrim:
> I'm ambivalent here: it would be nice to test against GHC 9.2 RC, but it means we'd have to update CI config once again when GHC 9.2.1 is out. 

That is a valid concern.  But you would have to update CI anyway when GHC 9.2.1 is released, so you do not lose anything by testing with the RC now.

Here are the steps to take after 9.2.1 is out:
1. wait for `haskell-ci` to support `9.2.1`
2. change the `tested-with` field in `text.cabal` from `9.2.0.20210821` to `9.2.1`
3. run `haskell-ci regenerate`
4. commit, push

The same steps are needed when you skip rc1, except that in step 2 you _add_ rather than _change_.

The rationale for including CI for 9.2 RC is that those that depend on `text` are more confident to bump to `text-2.0` in connection with 9.2.1 support.  (At least, this was the reason for me to open this PR.)

Bodigrim (approved):
All right, if it improves users' confidence in `text-2.0`, let's merge it early.

Issue #368, opened 2021-09-25
Please, add the subsequence predicate function


Anton-Latukha:
The [subsequence](https://en.wikipedia.org/wiki/Subsequence) predicate function use example is the textual fuzzy search.

In the most basic implementation, fuzzy search == subsequence search, as a defining feature of it. Examples of subsequence search being the main feature of fuzzy search are several I know of, the closest to me would be Emacs `helm` `fuzzy-find`, which for file name search does plain subsequence search to be fast.

The `isSubsequenceOf` exists in [`Data.List`](https://hackage.haskell.org/package/base-4.15.0.0/docs/src/Data-List.html#isSubsequenceOf).

To implement textual fuzzy search, the Text-specialized function needs to exist, so while being at it, would try to submit it.

PR #369, opened 2021-09-25
Text: add fun isSubsequenceOf

Anton-Latukha:
Closes report: #368

Function to detect ordered subsequences in the content.


src/Data/Text.hs
```diff
@@ -1877,6 +1878,22 @@ isInfixOf needle haystack
...
+-- 'True' iff the second is a subsequence of the first.
+-- (characters of the second argument appear in same sequential order in
+-- the first, to say if second argument can be derived by deleting some
+-- or no elements from the first).
+isSubsequenceOf :: Text -> Text -> Bool
+isSubsequenceOf tf sf
+  | length sf > length tf = False
+  | otherwise = subseqOf tf sf
+ where
+  subseqOf t s
```
Bodigrim, line 1891: I strongly suggest using `uncons` instead of `null` and `unsafeHead`

Bodigrim (commented):
@Anton-Latukha nice! But could we possibly squeeze out more performance by operating on underlying buffers? I imagine fuzzy search needs to be as fast as possible.

Anton-Latukha:
I currently do not have enough knowledge of low-level and high performance Haskell. But would learn and try to improve.

Currently, next I would cover the supplementary quality required: tests, benchmarks and things that good library piece requires. Having that would look into how further optimization.

Thank you for your work and reviews.

src/Data/Text.hs
```diff
@@ -1877,6 +1879,29 @@ isInfixOf needle haystack
     | otherwise          = not . L.null . indices needle $ haystack
 {-# INLINE [1] isInfixOf #-}
 
+-- | The 'isSubsequenceOf' function takes two 'Text's and returns
+-- 'True' iff the second is a subsequence of the first.
```
Lysxia, line 1883: This looks backwards. ``first `isSubsequenceOf` second`` should be testing whether `first` is a subsequence of `second`.

src/Data/Text.hs
```diff
@@ -1877,6 +1879,29 @@ isInfixOf needle haystack
...
+-- (characters of the second argument appear in same sequential order in
+-- the first, to say if second argument can be derived by deleting some
+-- or no elements from the first).
```
Lysxia, line 1886: It seems worth mentioning in the doc that this is useful for fuzzy search.

src/Data/Text.hs
Anton-Latukha: Well, this is how at least Lambda Calculus works.
So compiler sees where and how it can reuse the higher-order function & its memory.

The more constant and biggest impact thing goes first to cache it/cache precompiled specialized function to it, store what is needed in faster machine memory.

The other way is actually imperative style, which people got accustomed to.

That is why I talk that `lookup request db` is ... It is if we want to cache the same large request to search through a large number of small databases. It is illogical from the lambda calculus/System F design. But there were internal GHC work trying to mitigate the optimization of order of arguments, but its results, well *afaik* it worked good only for 2 argument functions.

While caching the db because it is the main data source and then number of different requests on it:
```haskell
  -- instead of:
  resCur = fmap (\ r -> lookup r db) requests -- lets use side-car lambda expression to fix lambda expression 
  -- or
  resCur = requests <&> (`lookup` db) -- lets use more exotic function to fight the natural language control flow, so we swap places of the fundamental operation in FP (function application), as read/write Haskell as if we do shell scripting, & and hide the lambda fix
  -- or
  resCur = requests ?? (`lookup` db) -- lets use lenses to fix the order of arguments in the lambda expression
  
  -- while all that time the solution was:
  res = fmap (lookup db) requests -- lets just use the proper order of arguments for the lambda expression language. As a result of it - there would be 1 major way to write the expression & at the same time it would be the most concise & elegant way.
  
  -- it is not hard to guess which of the above are easier to optimize for GHC
  
  -- & so
  conveyor =
    (<processing> . zip) (isSubsequenceOf doc <$> subsequences) subsequences  -- as expression - completely tail-optimized and so compiled-in lazy function to process different subsequences through some huge doc
```
it respects the Haskell, Haskell Core, System F, Lambda Calculus designs, therefore theory gets used in full on it, it is easier for the Haskell compiler to optimize this code, & so produce more optimized code, but also in itself code is concise & elegant.

In reality, it is debatable either the `carrier data should be searched for a pattern`, or the `pattern should be searched in the carrier data`, linguistically the first seems a simpler way of expressing instructions, but in Haskell, it is proper (by the lang design) to read it right to left.

It is just because people are accustomed to see & read the Haskell, lets say - not in `O<-I` order it is designed in, or in data processing order in the lambda calculus (the same argument), but in some imperative-styly order, we have this thing. To make things harder - language provides pretty powerful `do` notation that reverses the (tails to the beginning) FP control flow into partially imperative style (beginning to tails), reinforcing the puzzlement.

Lets say this is WIP code, it is done currently in the way that is easy to understand, transform & optimize, main purpose is to gradually refactor it into an optimized version, then it would be transformed into the resulting final code. If there would be requirements to flip the arguments - Ok. *afaik* GHC should be able to handle that.
 
But in GHC it (argument order optimization analysis) works solidly only for functions with 2 arguments. Someone needs to start solving the old mistakes somewhere, that is what I try to do. Higher-Order Function should be specialized for, saturated with the carrier data which is shared/reused between a lot of functions that do a lot of different requests to it.

Lysxia: The order of arguments does not impact performance at all here. The order of argument's I would expect is that `first `isSubsequenceOf` second` matches the English specification "`first` is a subsequence of `second`". Furhtermore this would be consistent with the order of arguments of `isInfixOf`.

Maybe you're thinking of a future implementation where you might precompute a state machine given the subsequence, but 1) this is not it, 2) then it could be called `isSupersequenceOf` instead.

Boarders:
I didn't test this (and so I wouldn't trust it to actually be correct!), but just to give you a rough idea of how a more efficient implementation might look, something like this might be a good starting point:

```haskell
isSubsequenceOf :: Text -> Text -> Bool
isSubsequenceOf (Text needle o1 l1) (Text target o2 l2) = runST $ go o1 o2
  where
    go :: forall s . Int -> Int -> ST s Bool
    go i _ | i >= l1 = return True
    go _ j | j >= l2 = return False
    go !i !j = do
      let !(Iter c d) =  iterArray needle i
      let !(Iter c' d') = iterArray target j
      if c == c'
        then go (i + d) (j + d')
        else go i (j + d')
```

There is plenty of good material to steal from in the rest of the library too if you want to see how to work with the underlying array. Getting tests and benchmarks going before figuring this out would be great though!

src/Data/Text.hs
Anton-Latukha: I've made a list of arcticles, but the text got lost.

You do not know about this problem, because it is already very old Computer Science / FP problem that is so scientifically trivial on its face of requiring to solve it & and developed, so it does not make hype splashes anymore.

Articles start from 1970 of data control flow analysis & SPJ has quite a number of articles, 2019 is particularly rich in references to other information. Guys & SPJ worked numerous times on arity raising optimization systems in GHC, articles related to argument ordering optimization, old and newer currying optimization systems, "Manticore" and so on and so on.

Terms that directly influenced & try to solve argument passing ordering:
* `data-flow alanysis`
* `function partitioning`
* `partial application`, currying & uncurring.
* `free variables and problems of inlining
* `demand analysis`, `specializalization`, `inlining` & `instantiaton`
* curring/uncurrying optimization
* `arity raising` & its optimization approaches and techniques
* `G-mashine`, `Spineless Tagless G-Machine` and its mapping to hardware.

The scientific community does not refer to preferable ordering of arguments, because for 50 years it tries to tackle the different optimization algorithms for it, and as SPJ articles show - the problem is still fresh.

Scientists do not talk about ordering, because the recommendations are very old & obvious to them, and as it is not their responsibility to order arguments properly in the functions.

There are not only articles, there are books that refer to all of this & GHC.

If one reads carefully "Haskell Programming from First Principles" it also covers the question of a canonical ordering of arguments for Haskell that align with currying/partial application/inlining and books gives recommendations on them. There is an optimal order to argument passing that help the data flow analysis, demand analysis & reuse.

The problem starts when people think that ordering of arguments does not matter to the code & result. People implement the wrong order of arguments and then everyone requires to follow them.

The type/ordering that does:
```haskell
res = fmap (lookup db) requests
```
is a proper order, from the style of the code & the science behind the language.

But, if to phrase & argument it as respecting the language design, preserving & propagating the best practices - does not matter, but what matters is what mistakes was fossilized previously - honestly that offends me as a kind of some programmer.

As an educated person, I can not accept the "it was this way before, so should be the same" arguments. Those are one of the main archetypes of a fallacious argumentation in a debate, the most stong one that perpetuates the errors, it is literally the first phrase mentioned in the [articles 1](https://rationalwiki.org/wiki/List_of_cognitive_biases) [2](https://www.boardofinnovation.com/blog/16-cognitive-biases-that-kill-innovative-thinking/) on cognitive biases we have.

But if to admit the problem: "I know the old ordering is wrong, but lets introduce new function with that wrong ordering, GHC in 2 arguments it does not matter, we can mention that canonical ordering in Haskell is different, lets just make it again to make it seamless to people as other functions made the same mistake", it is nice to phrase it in that way to negotiate with a person to step over the science & knowledge and purposefully do the wrong thing that perpetuates a bad code style. Because sometimes it indeed is needed, but the need & how to mitigate the problem is determined during the dialectic debate.

As I see it - `isSubsequenceOf` is a pretty rare function, the main use of it is for textual search. So despite the error in the earlier implementation in `[]`, the `text` implementation would be a more impactful example, and so it can set a proper ordering here, list is notorious and so on and so on, avoid `[]` asap ... and so the `bytestring` & another can/would follow `text` implementation. The base would be rehashed in the future and so `text` with proper ordering can indicate one more problem to fix in the `base` during transitions & breakages there.

Or at least the other function can be created & the `isSubsequenceOf` is an inalienable synonym for a proper function.

Forming these messages takes time, I think this is the most pinpoint question in the flow, would be better to spend the time on writing the code, and I think we are interested in the quality code.

This message between the lines proposed the solution: if maintainers ask - Ok, would do that ordering, but lets make `isSubsequenceOf` a synonym to the proper function and mention that in the doc, which would:
* help the compiler optimization, as flip alias would be annihilated during compilation - the alignment of the currying in Core would be right.
* give people function with proper ordering, the elegance of its use would nudge people towards the proper ordering.
* mention in docs that `isSubsequenceOf` is a synonym, to explicitly politely nudge people towards proper practice, even if without mentioning why.
* In the new function mention that its argument ordering is canonical for Haskell.
* as a result of all that - the reverse compatibility is preserved, but a new function is introduced and encouraged to be used, and so in the future other libraries would propagate the new function name.

Then we are arriving at the most difficult question - how to name it.

Bodigrim: I feel that we digress here. The [original ticket](https://github.com/haskell/text/issues/368) has a pretty solid motivation: there is [`Data.List.isSubsequenceOf`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:isSubsequenceOf), it is desirable to have a counterpart working over `Text`. However, swapping arguments `Data.Text.isSubsequence` to an opposite of `Data.List.isSubsequenceOf` directly contradicts this goal. 

Currently all `isFooBarOf` functions in core packages take a needle first and a haystack second. If there is an interest to change this convention, let's discuss it separately (and probably not even in `text` issue tracker). FWIW I think that even from a theoretical viewpoint it is better to keep a needle first: fuzzy search compares a fixed needle against thousands of file names.

Further, if there is a desire to add `isSupersequenceOf` or any other function to the public API, let's discuss it in an issue first. 

In the meantime I suggest we stick to the convention and make `isSubsequenceOf` to take a needle first.

Anton-Latukha: Yes.

Argument orders are a thing that I have an interest in. It is one of the questions I get railed up about quickly. And a think I want to make a general practical experience and formulate the ordering practices more solidly.

Yesterday I understood that sometimes it is a too aerial thing to deliver it nicely.

Also, I understood that my such reactions are indeed my problem, & that I should work on in the other way compared to how I was working before on it, it is indeed an emotional problem of mine.

But a dive into functions & arguments to them is a great topic to make a dive.

The [Making a Faster Curry with Extensional Types](https://www.microsoft.com/en-us/research/uploads/prod/2019/07/arity-haskell-symposium-2019.pdf) is a great article that touches on some main FP questions & at the same time a great index article.

src/Data/Text.hs
```diff
@@ -1881,6 +1883,43 @@ isInfixOf needle haystack
...
+    on f uncons t s
+   where
+    f :: Maybe (Char, Text) -> Maybe (Char, Text) -> Bool
+    f _ Nothing = True
+    f Nothing _ = False
+    f (Just (tc,ts)) (Just (sc,ss)) =
+      subseqOf ts $
+        if tc == sc
+          then s
+          else ss
```
axman6-da, line 1921: Is there a good reason not to just use 

```haskell
isSubsequenceOf needles haystack = List.isSubsequenceOf (T.unpack needles) (T.unpack haystack)
```
This code is doing exactly the same work, since `T.unpack` (as far as I understand) is just repeated calls to `uncons`. I can't see how there's be any performance here (though I could see an argument to using an optimised implementation in C if there's a more efficient algorithm that takes advantage of the UTF-8 encoding).

src/Data/Text.hs
```diff
@@ -1881,6 +1883,43 @@ isInfixOf needle haystack
...
+-- to find and returns 'True' iff the second argument is a subsequence
+-- of the first.
+--
+-- "Subsequence" used in the meaning of: characters of the second argument
+-- appear in same sequential order in the main data, to say second argument can
+-- be derived by deleting some (any) or no elements from the first.
+--
+-- Examples:
+--
+-- >>> isSubsequenceOf "1234567" "1356"
```
axman6-da, line 1898: Aren't the arguments her backwards? Generally `isInfixOf` etc. are used infix:

```haskell
>>> "ello" `isInfixOf` "Hello, world!"
True
```
but this would be
```haskell
"1234567" `isSubsequenceOf` "1356"
```
which should be False, if you read it in English. It should be `isSubsequenceOf needles haystack` - `"1234567"` _is not_ a subsequence of `"1356"`.

tests/Tests/Properties/Substrings.hs
```diff
@@ -231,6 +233,47 @@ tl_isSuffixOf s   = L.isSuffixOf s`eqP` TL.isSuffixOf (packS s)
 t_isInfixOf s     = L.isInfixOf s `eqP` T.isInfixOf (packS s)
 tl_isInfixOf s    = L.isInfixOf s `eqP` TL.isInfixOf (packS s)
 
+-- | Generator for substrings that keeps the element order.
+-- Aka: "1234567890" -> "245680"
+genOrdSubseq :: T.Text -> Gen T.Text
+genOrdSubseq txt =
+  T.pack . transform <$> genTransformMap
```
axman6-da, line 240: I can't understand what this implementation is actually doing, but is this not all we need? 

```suggestion
genOrdSubseq txt = 
  T.pack . map fst . filter snd . zip (T.unpack txt) <$> vectorOf (T.length txt) arbitrary :: Gen [Bool]
```
We're just trying to select a random sample of elements from the given string in order right? Are there any other properties we need from this? Trying to ensure that the result is non-empty isn't too hard, you just generate two lists of Bools:
```haskell
```suggestion
genOrdSubseq txt = do
    firstLen <- choose (1,T.length txt - 2)
    prefix <-  vectorOf firstLen arbitrary
    suffix <- vectorOf (T.length txt - firstLen - 1) arbitrary
    pure $ T.pack . map fst . filter snd . zip (T.unpack txt) $ prefix ++ True : suffix
```
(not sure if I have the maths right, but I'm trying to make a list of `length txt` which has a `True` somewhere in the middle)

axman6-da (commented):
Happy to see the function in the library, but there seems to be some issues with this implementation. 

If we had efficient implementations of `findCharIndex :: Char -> Text -> Maybe Int` then we could make this function quite fast in C on the underlying utf-8 representation.

axman6-da:
uh, sorry, I didn't see others' reviews when I was writing mine, and it looks like most of my comments have been mentioned already. Sorry for the duplicate review, but I hope I have something to add nonetheless.

Lysxia:
I would accept the implementation mentioned in an earlier comment:

```
isSubsequenceOf needles haystack = List.isSubsequenceOf (T.unpack needles) (T.unpack haystack)
```

If you want to stick to the current implementation,

- the order of argument should be the same as `Data.List.isSubsequenceOf`
- the auxiliary `subseqOf` should avoid doing redundant `uncons` when the needle doesn't change. This can be done by specializing it to a non-empty needle (represented as a `Char` and `Text` argument

Compared to using `List.isSubsequenceOf`, a well optimized implementation would have the advantage of not allocating intermediate data structures (I don't think you need any unsafe primitives, GHC's worker-wrapper + inlining should be sufficient).

PR #371, opened 2021-09-28
Freshen up README

Bodigrim:


Boarders (approved):
Good update, thanks! Is there a reason to prefer recommening https over ssh?

Bodigrim:
Currently README features one ssh and one https example, so I just unified both to be https. I do not have a strong preference, but cloning by https does not require users to setup ssh keys, and I also find URLs a bit more natural.

Lysxia (approved):

PR #373, opened 2021-09-29
Optimize handling of Text literals

Bodigrim:
Currently a literal `"foo" :: Data.Text.Text` (with `OverloadedStrings` enabled) compiles to an opaque function call:

```haskell
-- RHS size: {terms: 1, types: 0, coercions: 0, joins: 0/0}
foo1 :: Addr#
foo1 = "foo"#

-- RHS size: {terms: 2, types: 0, coercions: 0, joins: 0/0}
foo :: Text
foo = unpackCString# foo1
```

https://github.com/haskell/text/blob/3488190f25e1ee5a5dea6b1a593d8a5819e76a1e/src/Data/Text/Show.hs#L54-L56

So every literal is streamed and unstreamed in runtime, and even the constructor `Text` is not known in compile time, preventing any worker-wrapper transformations. 

With this patch the same `"foo" :: Text` gives (for GHC 9.0+)

```haskell
-- RHS size: {terms: 1, types: 0, coercions: 0, joins: 0/0}
foo1 :: Addr#
foo1 = "foo"#

-- RHS size: {terms: 24, types: 33, coercions: 0, joins: 0/0}
foo :: Text
foo
  = runRW#
      (\ (s_s3O7 :: State# RealWorld) ->
         case newByteArray# 3# s_s3O7 of { (# ipv_i3zp, ipv1_i3zq #) ->
         case copyAddrToByteArray# foo1 ipv1_i3zq 0# 3# ipv_i3zp of s2#_i3C0
         { __DEFAULT ->
         case unsafeFreezeByteArray# ipv1_i3zq s2#_i3C0 of
         { (# ipv2_i3Ca, ipv3_i3Cb #) ->
         Text ipv3_i3Cb 0# 3#
         }
         }
         })
```

Not only the constructor, but also offset and length are now known in compile time, and the only runtime operation is memory copy. GHC could even void allocation of `Text` constructor, if things get inlined and worker-wrapper transformation kicks in successfully.


chessai (approved):

chessai:
Beautiful!

Bodigrim:
My approach leaves unhandled surrogate pairs, which can appear in `Addr#`, but should not appear in `Text`. Will fix by the end of the week.

phadej:
This is incorrect. Try with `"\0"`.

Bodigrim:
@phadej as witnessed by CI `"\0"` works fine, the patch handles `\xc0\x80` correctly.

src/Data/Text/Show.hs
```diff
@@ -46,18 +54,56 @@ unpack ::
...
+        = A.unsafeWrite marr dstOff w8 >> go (srcOff + 1) (dstOff + 1)
+  actualLen <- go 0 0
+  A.shrinkM marr actualLen
+  arr <- A.unsafeFreeze marr
+  return $ Text arr 0 actualLen
+{-# INLINE unpackCString# #-}
+
+-- | /O(n)/ Convert a literal string into a 'Text',
+-- counterpart to 'GHC.unpackCString#'.
+--
```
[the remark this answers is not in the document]
phadej: ... and that it doesn't handle zero byte, as there is no length given.

Lysxia:
Where can I read about the "modified UTF-8" encoding of string literals?

phadej:
@Lysxia 

> Where can I read about the "modified UTF-8" encoding of string literals?

AFAIK it's Java invention: https://docs.oracle.com/javase/6/docs/api/java/io/DataInput.html#modified-utf-8 to avoid null byte in strings.

Bodigrim:
This is ready for a final review now.

src/Data/Text/Show.hs
```diff
@@ -46,18 +54,65 @@ unpack ::
 unpack = S.unstreamList . stream
 {-# INLINE [1] unpack #-}
 
--- | /O(n)/ Convert a literal string into a 'Text'.
---
--- This is exposed solely for people writing GHC rewrite rules.
+-- | /O(n)/ Convert a null-terminated
+-- <https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8 modified UTF-8>
```
Lysxia, line 58: There seems to be a difference from Modified UTF-8. Aren't 4-byte UTF-8 codepoints still represented as themselves (i.e., in 4 bytes) in Haskell literals (which is why we can afford to erase surrogates), whereas Java's Modified UTF-8 actually must encode them as surrogate pairs?

Bodigrim: Yes, Haskell version of Modified UTF-8 represent 4-byte codepoints transparently. "**All known** Modified UTF-8 implementations also treat the surrogate pairs as in CESU-8" is false, Wiki should say "Most".

Lysxia (approved):
LGTM

Bodigrim:
Doctests are broken even without any meaningful changes to haddocks. Let's wait for #376 to land first.

Bodigrim:
CI is green finally.

PR #383, opened 2021-11-13
Add s390x and ppc64le CI builds

Bodigrim:
It's important to run builds on a big-endian architecture.

### DIFF
--- a/tests/Tests/QuickCheckUtils.hs
+++ b/tests/Tests/QuickCheckUtils.hs
@@ -248,11 +248,13 @@ write_read unline filt writer reader nl buf ts = ioProperty $
     t = unline . map (filt (not . (`elem` "\r\n"))) $ ts
 
     act = withTempFile $ \path h -> do
+            IO.hSetEncoding h IO.utf8
             IO.hSetNewlineMode h nl
             IO.hSetBuffering h buf
             () <- writer h t
             IO.hClose h
             bracket (IO.openFile path IO.ReadMode) IO.hClose $ \h' -> do
+              IO.hSetEncoding h' IO.utf8
               IO.hSetNewlineMode h' nl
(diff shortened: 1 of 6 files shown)


tests/Tests/QuickCheckUtils.hs
```diff
@@ -49,7 +49,7 @@ import qualified Data.Text.Lazy as TL
 import qualified System.IO as IO
 
 genWord8 :: Gen Word8
-genWord8 = chooseAny
+genWord8 = choose (minBound, maxBound)
```
phadej, line 52: I don't understand the commit message:  the bounds are `QuickCheck >= 2.14.1 && < 2.15`?

Bodigrim: The thing is that in emulated environment I have no resources to build a proper QuickCheck and must rely on whatever is `libghc-quickcheck2-dev` is. For Ubuntu 20.04 it is [2.12.6.1](https://packages.ubuntu.com/focal/libghc-quickcheck2-dev). Since `ghc --make` does not even look at cabal file, I can get away without changing bounds.

Let me do it in a proper way, with a CPP switch.

Bodigrim: Ah, right, I do not need CPP at all, just import `chooseAny` from `Test.QuickCheck.Gen` instead of `Test.QuickCheck`.

Bodigrim: Thanks, this is fixed now.

Bodigrim:
@Boarders @parsonsmatt @Lysxia could someone please take a look at this? The workflows are similar to https://github.com/haskell/bytestring/pull/436 and I tested them on several other projects.

Lysxia (approved):

PR #384, opened 2021-11-14
Regenerate case mappings for Unicode 14.0 with matching GHC 9.2

Bodigrim:
#372 is incomplete: it's important to regenerate mappings using a version of GHC, which supports a respective version of Unicode. GHC 9.2 supports Unicode 14.0, so now case mappings are correct.

### DIFF
--- a/src/Data/Text/Internal/Fusion/CaseMapping.hs
+++ b/src/Data/Text/Internal/Fusion/CaseMapping.hs
@@ -1064,6 +1064,7 @@ upperMapping = \case
   '\x2c5c'# -> unI64 11308
   '\x2c5d'# -> unI64 11309
   '\x2c5e'# -> unI64 11310
+  '\x2c5f'# -> unI64 11311
   '\x2c61'# -> unI64 11360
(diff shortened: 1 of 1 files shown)


Bodigrim:
https://github.com/haskell/text/blob/fdc938628efe9ecf86e7d11128c4c7cc66d1071a/scripts/SpecialCasing.hs#L51
https://github.com/haskell/text/blob/fdc938628efe9ecf86e7d11128c4c7cc66d1071a/scripts/CaseFolding.hs#L44
Specifically, this code relies on `Data.Char.to{Lower,Upper,Title}` to determine whether a character has case inflection. If `base` version is not in line with Unicode version in http://unicode.org/Public/UNIDATA/SpecialCasing.txt and http://unicode.org/Public/UNIDATA/CaseFolding.txt, this produces unexpected results and causes 
https://github.com/haskell/text/blob/d615c996d480d527c468a72af8750b766da4ad02/tests/Tests/Properties/Text.hs#L118-L129
to fail.

Lysxia (approved):

PR #387, opened 2021-11-21
Undeprecate decodeASCII

Bodigrim:
With UTF8 representation there is no longer a reason to deprecate `decodeASCII`: all three `decodeASCII`, `decodeLatin1` and `decodeUtf8` have different trade offs. 

For example, `aeson` can benefit from using `decodeASCII`, see https://github.com/haskell/aeson/blob/06114272b1889f27a9833d01747b3372c71bcbc4/src/Data/Aeson/Parser/Internal.hs#L345-L349 and discussion at https://github.com/haskell/aeson/issues/893#issuecomment-974404169.

### DIFF
--- a/benchmarks/haskell/Benchmarks/DecodeUtf8.hs
+++ b/benchmarks/haskell/Benchmarks/DecodeUtf8.hs
@@ -1,5 +1,4 @@
 {-# LANGUAGE ForeignFunctionInterface #-}
-{-# OPTIONS_GHC -fno-warn-deprecations #-}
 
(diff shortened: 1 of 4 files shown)


Lysxia (approved):

ethercrow:
> all three decodeASCII, decodeLatin1 and decodeUtf8 have different trade offs.

Maybe add the guidance on choosing between them to their haddocks?

Bodigrim:
@ethercrow done, thanks.

PR #389, opened 2021-11-21
Tests: Generate chunks that start in the middle of their buffers

Lysxia:
Following up on #386 we can put a `drop` in the `Arbitrary` instance to generalize other tests to check that offsets are handled correctly.


Bodigrim (approved):
Nice! Test failures for `splitOn` are caused exactly by the issue, which #386 resolves, so I suggest we merge it first.

Bodigrim:
Thanks!

PR #390, opened 2021-11-22
Make streamDecodeUtf8 stricter

Bodigrim:
Discovered this while testing `conduit` with `text-2.0`.

In `text-1.2.5.0`
```
> Data.Text.Encoding.streamDecodeUtf8 (Data.ByteString.pack [0x81]) `seq` ()
*** Exception: Cannot decode byte '\x81': Data.Text.Internal.Encoding.streamDecodeUtf8With: Invalid UTF-8 stream
```

But in `text-2.0`
```
> Data.Text.Encoding.streamDecodeUtf8 (Data.ByteString.pack [0x81]) `seq` ()
()
```

This can be fixed by forcing `txt` in 
https://github.com/haskell/text/blob/922d9cc7c537d55e3be2dba2343ef43e25493ace/src/Data/Text/Encoding.hs#L365-L369

But I decided to fix it in a more principled way by making fields of `Decoding` strict.

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -328,7 +328,7 @@ decodeUtf8With2 onErr bs1@(B.length -> len1) bs2@(B.length -> len2) = runST $ do
 -- | A stream oriented decoding result.
 --
 -- @since 1.0.0.0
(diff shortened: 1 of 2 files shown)


Bodigrim:
@Boarders @parsonsmatt @Lysxia This is ready for review.

PR #393, opened 2021-12-05
Make encodeUtf8BuilderEscaped faster

Bodigrim:
`aeson` benchmarks signalled (https://github.com/haskell/aeson/issues/893#issuecomment-986215880) that `encodeUtf8BuilderEscaped` got slower after UTF8 transition. This PR makes it 2-3x faster:

```
All
  EncodeUtf8
    non-ASCII
      Text/encodeUtf8BuilderEscaped: OK (3.96s)
        1.90 ms ±  53 μs, 46% faster than baseline
    ASCII
      Text/encodeUtf8BuilderEscaped: OK (10.32s)
        629  μs ± 9.0 μs, 70% faster than baseline
```

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -90,7 +90,7 @@ import qualified Data.ByteString.Builder as B
 import qualified Data.ByteString.Builder.Internal as B hiding (empty, append)
 import qualified Data.ByteString.Builder.Prim as BP
 import qualified Data.ByteString.Builder.Prim.Internal as BP
-import Data.Text.Internal.Encoding.Utf8 (utf8LengthByLeader, utf8DecodeStart, utf8DecodeContinue, DecoderResult(..))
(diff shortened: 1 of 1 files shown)


Bodigrim:
CC @Lysxia @Boarders @parsonsmatt

src/Data/Text/Encoding.hs
```diff
@@ -474,33 +474,18 @@ encodeUtf8BuilderEscaped be =
           -- is smaller than 8, as this will save on divisions.
           | otherwise        = return $ B.bufferFull bound op0 (outerLoop i0)
           where
-            outRemaining = (ope `minusPtr` op0) `div` bound
+            outRemaining = (ope `minusPtr` op0) `quot` bound
```
Lysxia, line 477: This doesn't change anything, does it?

Bodigrim: Both operands are positive, so the result is unchanged. `quot` usually optimizes better, because it matches CPU instruction for signed division, while `div` requires branches. This is a micro-optimization indeed.

Lysxia (approved):

PR #394, opened 2021-12-05
Faster fallback options when simdutf is not available

Bodigrim:
Currently when `simdutf` is disabled, we fallback to the slowest mode of operation: decode every byte into code points, then encode each code point back. This is wasteful and extremely (100x) slow. This PR provides two fallback options:

* If [`bytestring-0.11.2.0`](https://hackage.haskell.org/package/bytestring-0.11.2.0) is available, use [`Data.ByteString.isValidUtf8`](https://hackage.haskell.org/package/bytestring-0.11.2.0/docs/Data-ByteString.html#v:isValidUtf8).
* Otherwise use Haskell native implementation, but in validation-only mode, which is 10x faster than native decoding/encoding.

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -79,9 +79,6 @@ import Data.Text.Show as T (singleton)
 import Data.Text.Unsafe (unsafeDupablePerformIO)
 import Data.Word (Word8)
 import Foreign.C.Types (CSize(..))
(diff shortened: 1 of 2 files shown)


Bodigrim:
CC @Lysxia @Boarders @parsonsmatt

Lysxia (approved):

PR #397, opened 2021-12-17
Remove INLINE isPrefixOf

### LINKED ISSUE #351: `lines` can't be inlined
Building `text` with a debugging GHC, it reports:

```
libraries/text/src/Data/Text.hs:1699:1: warning:
    INLINE binder is (non-rule) loop breaker: lines
```

Indeed `lines` is recursive and can't be inlined. We should either remove the `INLINE` pragma or use a local recursive function.

Similarily:

```
libraries/text/src/Data/Text/Lazy.hs:1564:1: warning:
    INLINE binder is (non-rule) loop breaker: isPrefixOf
```

Lysxia:
Closes #351 (the other pragma for `lines` was already removed at some point)

### DIFF
--- a/src/Data/Text/Lazy.hs
+++ b/src/Data/Text/Lazy.hs
@@ -1472,7 +1472,6 @@ isPrefixOf (Chunk x xs) (Chunk y ys)
         (yh,yt) = T.splitAt lx y
         lx = T.length x
         ly = T.length y
-{-# INLINE [1] isPrefixOf #-}
 
 -- | /O(n)/ The 'isSuffixOf' function takes two 'Text's and returns
 -- 'True' iff the first is a suffix of the second.


Boarders (approved):

Bodigrim:
Could someone please explain how to reproduce analysis in #351? Which GHC flags should I use?

Lysxia:
> Could someone please explain how to reproduce analysis in #351? Which GHC flags should I use?

CC @hsyl20  IIUC it's only available on a debugging build of GHC.

hsyl20:
> CC @hsyl20 IIUC it's only available on a debugging build of GHC.

The warning is generated by core-lint, so perhaps using `-dcore-lint` is enough. No need to use a debugging compiler.

Bodigrim (approved):

PR #399, opened 2021-12-25
Special-case openbsd c++ libraries

### LINKED ISSUE #398: extra-libraries for c++ are incomplete
`text.cabal` has this fragment which likely belongs somewhere in cabal:
```
  if os(windows)
    if arch(x86_64)
      extra-libraries: stdc++-6 gcc_s_seh-1
    else
      extra-libraries: stdc++-6 gcc_s_dw2-1
  else
    if os(darwin)
      extra-libraries: c++
    else
      extra-libraries: stdc++
```

As it stands, the fragment is incomplete and some clang-based operating systems need slightly different libraries for this to work. E.g. ghci in HEAD complains about missing libraries [somewhat incomprehensibly](https://gitlab.haskell.org/ghc/ghc/-/issues/20870).

I'll try to cook up a patch which solves the immediate problem, but it's likely not at the right level.

blackgnezdo:
They may be typical for clang-based systems, but I'm not sure what
that list is.

Fix #398

Related to https://gitlab.haskell.org/ghc/ghc/-/issues/20870

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -95,7 +95,7 @@ library
         extra-libraries: stdc++-6 gcc_s_seh-1
       else
         extra-libraries: stdc++-6 gcc_s_dw2-1
-    elif os(darwin)
+    elif os(darwin) || os(freebsd)
       extra-libraries: c++
     elif os(openbsd)
       extra-libraries: c++ c++abi
--- a/text.cabal
+++ b/text.cabal
@@ -95,11 +95,12 @@ library
         extra-libraries: stdc++-6 gcc_s_seh-1
       else
         extra-libraries: stdc++-6 gcc_s_dw2-1
+    elif os(darwin)
+      extra-libraries: c++
+    elif os(openbsd)
+      extra-libraries: c++ c++abi
     else
-      if os(darwin)
-        extra-libraries: c++
-      else
-        extra-libraries: stdc++
+      extra-libraries: stdc++
 
   exposed-modules:
     Data.Text


Bodigrim:
Thanks! Could you please add the same incantation for FreeBSD and check whether it helps? There is a CI job for FreeBSD, e. g., https://cirrus-ci.com/task/4653704308588544

blackgnezdo:
> Thanks! Could you please add the same incantation for FreeBSD and check whether it helps? There is a CI job for FreeBSD, e. g., https://cirrus-ci.com/task/4653704308588544

I added a commit. The darwin branch seemed like the right place as libc++api seems to be there on FreeBSD.

Bodigrim:
Nice, FreeBSD build completes without warnings now. Since we do not have OpenBSD CI, is there any explanation you could add as a comment except "that's how it works"? How do you know which systems need `libc++abi` and which don't?
(Sorry, if it's a stupid question, I do not know much about this stuff)

blackgnezdo:
> Nice, FreeBSD build completes without warnings now. Since we do not have OpenBSD CI, is there any explanation you could add as a comment except "that's how it works"? How do you know which systems need `libc++abi` and which don't? (Sorry, if it's a stupid question, I do not know much about this stuff)

Not having `libc++abi` caused a bunch of undefined symbols, like `__cxa_pure_virtual`. I don't know why different systems end up with these in different libraries. Here's what they look like:

On FreeBSD bum.src.cx 13.0-STABLE:
```
% nm /usr/lib/libc++.a |grep __cxa_pure_virtual
                 U __cxa_pure_virtual
                 U __cxa_pure_virtual
                 U __cxa_pure_virtual
0000000000000000 T __cxa_pure_virtual
```

On OpenBSD 7.0:
```
% nm /usr/lib/libc++.a |grep __cxa_pure_virtual
         U __cxa_pure_virtual
         U __cxa_pure_virtual
         U __cxa_pure_virtual
% nm /usr/lib/libc++abi.a  |grep __cxa_pure_virtual
         U __cxa_pure_virtual
00000000 T __cxa_pure_virtual

```

Bodigrim (approved):

blackgnezdo:
Any chance this can get merged/released so I can start pestering GHC folks to including the fix?

Bodigrim:
I'd like to hear from other maintainers before merge (CC @Boarders @Lysxia). 
I'm reluctant to make a separate release for this fix only, let's wait until we accrue something else (or learn that everything is just perfect ;).

Lysxia (approved):

Lysxia:
This PR looks good to me. I agree with Bodigrim that we should wait a bit before the next release. That seems especially indicated since 2.0 just came out. If any big issues come up as people upgrade, it would be nice to fix them in one go.

Bodigrim:
Thanks @blackgnezdo!

PR #400, opened 2021-12-28
Fix for macOS 10.13 (error: instruction requires: AVX-512 ISA)

vshabanov:
Hi, `text-2.0` doesn't compile on older macOS 10.13. It shows
```
     error: instruction requires: AVX-512 ISA
            vmovdqa64       LCPI2_0(%rip), %zmm0 ## zmm0 = [191,191,191,191,191, ...
     ...
```

While macOS 10.13 says it has clang 10, it actually has clang 6, and 10 is the version of Xcode.

Clang 6 does not enable proper -march flags for assembly modules which leads to the above compilation errors. It's fixed in clang 7. 

This patch fixes the issue by disabling AVX512 code for old macOS and clang <= 6.


cbits/measure_off.c
```diff
@@ -17,19 +17,38 @@
...
+  Disable AVX-512 instructions as they are most likely not supported
+  on the hardware running clang-6.
+*/
+#if (defined(__apple_build_version__) && __apple_build_version__ <= 10001145) \
+    || (defined(__clang_major__) && __clang_major__ <= 6)
+#define NO_AVX512
+#endif
+
+#if defined(__x86_64__) && !defined(NO_AVX512)
+#define USE_AVX512
```
Bodigrim, line 38: Thanks, @vshabanov, much appreciated! I have a tiny nit pick.

The name `USE_AVX512` is a bit misleading: it's not like we will definitely use AVX512 on any configuration, where `USE_AVX512` is defined. 

Could we please rework the PR to define `COMPILER_SUPPORTS_AVX512` macro and use it below as `#if defined(__x86_64__) && defined(COMPILER_SUPPORTS_AVX512)`?

phadej:
This will clash with #401. I'll leave it up to maintainers to resolve the conflicts as the patch in #401 is very trivial. (And there are tests).

cbits/measure_off.c
vshabanov: I've changed USE_AVX512 to COMPILER_SUPPORTS_AVX512.

Haven't done anything with #401 conflict yet as I don't know how to handle it better. If `__get_cpuid_count` is not supported then there's no point in having AVX512 functions as well. So we could either use `(__GNUC__ >= 6 || defined(__clang_major__))` to define COMPILER_SUPPORTS_AVX512 (which would make the macro badly named again) or we could just use it to `return false` in `has_avx512_vl_bw()`, which is probably fine.

Bodigrim: Second option please.

vshabanov: Done. Merged #401.

Bodigrim (approved):
LGTM, thanks!

PR #403, opened 2022-01-01
Fix static linking on Windows

Bodigrim:
Fixes https://gitlab.haskell.org/ghc/ghc/-/issues/20878
CC @Mistuke

### DIFF
--- a/.github/workflows/windows_and_macOS.yml
+++ b/.github/workflows/windows_and_macOS.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: ['windows-latest', 'macOS-latest']
-        ghc: ['8.10.5']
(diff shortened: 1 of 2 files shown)


Bodigrim:
I think I originally copied Windows incantation from https://github.com/haskell/double-conversion/blob/e4b7956d51fbb79ada95412519bd2a6a680def23/double-conversion.cabal#L74-L78 without investigating whether `gcc_s_seh-1` is needed. CI suggests that it's not.

Mistuke:
You should be able to use `stdc++` like on Linux. `stdc++-6` will find the shared library again.

Bodigrim:
@Lysxia @Boarders this is ready for review.

Lysxia (approved):

PR #404, opened 2022-01-08
Prohibit unbuildable Windows configurations due to GHC bugs wrt TemplateHaskell and C interaction

### LINKED ISSUE #385: text-2.0 candidate release
I believe current `master` is mature enough to publish a candidate release for user acceptance testing. If there are no other suggestions, I'll upload it on Friday.

Bodigrim:
Closes #385. 
@RyanGlScott, could you please take a look?

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -99,11 +99,11 @@ library
 
   -- Certain version of GHC crash on Windows, when TemplateHaskell encounters C++.
   -- https://gitlab.haskell.org/ghc/ghc/-/issues/19417
-  if flag(simdutf) && os(windows) && (impl(ghc == 8.0.1) || impl(ghc >= 8.8) && impl(ghc < 8.10.5) || impl(ghc == 9.0.1))
+  if flag(simdutf) && os(windows) && impl(ghc == 8.0.1 || >= 8.8 && < 8.10.5 || == 9.0.1)
     build-depends: base < 0
 
   -- For GHC 8.2, 8.6.3 and 8.10.1 even TH + C crash Windows linker.
-  if os(windows) && (impl(ghc >= 8.2) && impl(ghc < 8.4) || impl(ghc == 8.6.3) || impl(ghc == 8.10.1))
+  if os(windows) && impl(ghc >= 8.2 && < 8.4 || == 8.6.3 || == 8.10.1)
     build-depends: base < 0
 
   exposed-modules:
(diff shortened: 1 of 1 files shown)


text.cabal
```diff
@@ -97,6 +97,15 @@ library
     else
       extra-libraries: stdc++
 
+  -- Certain version of GHC crash on Windows, when TemplateHaskell encounters C++.
+  -- https://gitlab.haskell.org/ghc/ghc/-/issues/19417
+  if flag(simdutf) && os(windows) && (impl(ghc == 8.0.1) || impl(ghc >= 8.8) && impl(ghc < 8.10.5) || impl(ghc == 9.0.1))
+    build-depends: base < 0
+
+  -- For GHC 8.2, 8.6.3 and 8.10.1 even TH + C crash Windows linker.
+  if os(windows) && (impl(ghc >= 8.2) && impl(ghc < 8.4) || impl(ghc == 8.6.3) || impl(ghc == 8.10.1))
```
phadej, line 106: You can write

```cabal
  -- For GHC 8.2, 8.6.3 and 8.10.1 even TH + C crash Windows linker.
  if os(windows) && impl(ghc (>= 8.2 && < 8.4) || ==8.6.3 || == 8.10.1)
    build-depends: base < 0
```

i.e. use same version range syntax as in `build-depends`. That is more readable IMO.

Bodigrim: Ah, thanks, great!

RyanGlScott (approved):
I'm no longer able to construct a build plan with GHC 8.2.2, so as far as I'm concerned, this fixes the issue for me. Thanks!

Bodigrim:
@Boarders @Lysxia this is ready for review.

Lysxia (approved):

RyanGlScott:
Is it possible to revise `text-2.0` on Hackage to include these changes? Even after the release of `text-2.0.1` (which includes this fix), `cabal` will still include `text-2.0` in build plans involving GHC 8.2 on Windows, which breaks the build. See [here](https://ci.appveyor.com/project/RyanGlScott/text-show/build/job/bw7wd4s31h0jrcqb) for an example.

Bodigrim:
I think Hackage revisions cannot add conditionals, so the only option is to deprecate `text-2.0` altogether, which I do not feel fully justified at this point.

RyanGlScott:
> I think Hackage revisions cannot add conditionals

Darn, that's really unfortunate. Is there a `hackage-server` issue tracking this somewhere?

In the meantime, I suppose I'll just have to exclude GHC 8.2 from CI on Windows.

phadej:
> > I think Hackage revisions cannot add conditionals
> 
> Darn, that's really unfortunate. Is there a `hackage-server` issue tracking this somewhere?
> 

I cannot find one, but i'm pretty sure there was sometehing either on `hackage-server` or `hackage-cli` issue trackers.

(relatedly, if https://github.com/hackage-trustees/hackage-cli/pull/9 lib is split out into standalone lib, so it's actually shared between hackage-server and hackage-cli might increase people motivation to contribute things. AFAIK those are also the only revision related tests existing - which is not that great)

PR #405, opened 2022-01-08
Use official Docker images for arm64

Bodigrim:
Expected to fail because of known issues with GHC 9.2.1.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,15 +1,25 @@
 kind: pipeline
-name: arm64
+name: arm64-ghc9.0
 platform: { os: linux, arch: arm64 }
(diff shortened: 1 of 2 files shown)


Bodigrim:
All green with GHC 9.2.2. Ready for review. CC @Boarders @Lysxia

Lysxia (approved):
LGTM. But why is there one remaining failure in CI for 8.10

Bodigrim:
@Lysxia well, now we know :) GHC 8.10 on ARM is known to have issues, so I just removed the job and explicitly prohibited such configuration in cabal file.

PR #410, opened 2022-02-20
Fix 'iff' typo

AlistairB:
Just noticed a small typo.


chessai:
iff means "if and only if"
but one could argue it's unnecessary here

AlistairB:
Hmm interesting, I had not heard of iff before, which perhaps makes the docs a bit less accessible.

I would suggest either change it to be explicit as 'if and only if', or just make it 'if'. I'm not sure if there is a semantic difference here, so I'd prefer the latter.

Anyway, happy to do whatever (or to close) :)

chessai:
I think I agree with this PR, `iff` is less well-known and slightly confusing to those not in the know

Bodigrim:
For those who are acquainted with the meaning of "iff", a commit changing "iff" to "if" would look as an intentional downgrade and may raise questions. Please put "if and only if" instead.

AlistairB:
@Bodigrim I wasn't sure if you just wanted the commit message changed or not. I changed the text as well if that is what you meant?

Lysxia:
I prefer this more explicit version too.

Bodigrim (approved):

PR #413, opened 2022-03-04
Make Lift instances more efficient

TeofilC:
This patch allows the Lift instance to make use of the primitive bytes literals made available since template-haskell 2.16. 

It also adds some extra tests to make sure the instance works correctly with multi codepoint characters.

Let me know if you'd like some changes


Bodigrim:
@phadej could you please review?

TeofilC:
The CI failures seem to be due to the change in template-haskell 2.17 for the type of `lift` from `a -> Q a` to `Quasi m => a -> m a`. I'll push a fix

Bodigrim (approved):

TeofilC:
I've force pushed to squash the fixup commit

TeofilC:
Ah dear it seems like I dropped that commit by accident instead of squashing it. I'll fix it in a bit

Bodigrim:
I'll squash all commits any way, so do not worry about that.

Bodigrim:
Thanks!

PR #421, opened 2022-04-09
text-test-data (clone)

askeblad:


askeblad:
the `haskell/text-test-data/Makefile` might be overkill

the `haskell/text/README.md` file section could be written like this

```
cd text
git clone https://github.com/haskell/text-test-data benchmarks/text-test-data
cd benchmarks && bunzip2 -k *.bz2
```

if the `haskell/text-test-data/Makefile` is deleted from that repo

then the proposed `haskell/text-test-data/.gitattributes` would be edited to remove the following line

`Makefile linguist-detectable=false`

askeblad:
_cf_: https://github.com/haskell/text-test-data/pull/2

README.md
```diff
@@ -9,13 +9,14 @@ Please report bugs via the
 
 The main repo:
 
-```bash
```
Bodigrim, line 12: Could you please keep `bash` in place here and below? Alternatively, replace `bash` with `sh`, there is nothing bash-specific.

askeblad: `cd` now appears in colour

askeblad:
@Bodigrim 

`haskell/text`
`haskell/text-test-data` 

unless changes in https://github.com/haskell/text/pull/421#issuecomment-1094137281 are indicated

ready to merge

Bodigrim (approved):

askeblad:
Thanks!

PR #422, opened 2022-04-14
Update simdutf to 1.0.1, fix compiling for wasm32-wasi

TerrorJack:
This patch checks in simdutf 1.0.1 source code, which supports the CPP macro needed to make it work on wasm32-wasi. The platform-specific logic in cabal file is also adjusted accordingly.

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -96,6 +96,10 @@ library
       extra-libraries: c++ c++abi pthread
     elif os(windows) && impl(ghc > 9.3)
       extra-libraries: c++ c++abi
+    elif arch(wasm32)
+      cpp-options: -DSIMDUTF_NO_THREADS
+      cxx-options: -fno-exceptions
+      extra-libraries: c++ c++abi
     else
       extra-libraries: stdc++
 
--- a/text.cabal
+++ b/text.cabal
@@ -94,6 +94,8 @@ library
       extra-libraries: c++
     elif os(openbsd)
       extra-libraries: c++ c++abi pthread
+    elif os(windows) && impl(ghc > 9.3)
+      extra-libraries: c++ c++abi
     else
       extra-libraries: stdc++
 
@@ -177,7 +179,7 @@ library
     cpp-options: -DASSERTS
 
   -- https://gitlab.haskell.org/ghc/ghc/-/issues/19900
-  if os(windows)
+  if os(windows) && impl(ghc < 9.3)
     extra-libraries: gcc_s
 
   default-language: Haskell2010
--- a/simdutf/simdutf.cpp
+++ b/simdutf/simdutf.cpp
@@ -1,4 +1,4 @@
-/* auto-generated on 2021-07-29 10:43:28 -0400. Do not edit! */
+/* auto-generated on 2022-03-21 23:28:26 -0400. Do not edit! */
 // dofile: invoked with prepath=/Users/lemire/CVS/github/simdutf/src, filename=simdutf.cpp
 /* begin file src/simdutf.cpp */
 #include "simdutf.h"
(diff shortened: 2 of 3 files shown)


phadej:
```
% cabal check
Warning: These warnings may cause trouble when distributing the package:
Warning: 'ghc-options: -O2' is rarely needed. Check that it is giving a real
benefit and not just imposing longer compile times on your users.
Warning: The following errors will cause portability problems on other
environments:
Warning: Unknown architecture name 'wasm32'
Warning: Hackage would reject this package.
```

Bodigrim:
@TerrorJack could you please split this into three PRs? 

1. `simdutf` update itself could be merged immediately.
2. 5a43fec requires a CI job against sufficiently new GHC, which can probably be obtained via `ghcup`, CC @hasufell. 
3. `arch(wasm32)` is the most problematic, it would be better to keep a separate branch with it until Cabal is updated.

hasufell:
You can download a GHC from a branch pipeline, then we add it to downloads.haskell.org/~ghcup/unofficial-bindists, then add it to https://github.com/haskell/ghcup-metadata/blob/master/ghcup-prereleases-0.0.7.yaml and then you can enable it via `ghcup config add-release-channel <url>` in CI.

TerrorJack:
> simdutf update itself could be merged immediately.

Fair, I'll open a separate PR for simdutf update alone and close this one for the time being.

> https://github.com/haskell/text/commit/5a43fecb62e9769c4cf8b456c7b273901048ea72 requires a CI job against sufficiently new GHC, which can probably be obtained via ghcup

That commit was accidentally included in this PR, sorry about that; it exists on https://gitlab.haskell.org/ghc/packages/text/-/tree/wip/windows-clang, which I believe is a temporary workaround pushed by @bgamari but not opened a PR yet

> arch(wasm32) is the most problematic

Yes, hadrian had no problem with it but `cabal check` surfaced the problem. I will open a PR to cabal and hold the `arch(wasm32)` change for now.

TerrorJack:
Closed in favor of #424 for now.

Bodigrim:
> That commit was accidentally included in this PR, sorry about that; it exists on https://gitlab.haskell.org/ghc/packages/text/-/tree/wip/windows-clang, which I believe is a temporary workaround pushed by @bgamari but not opened a PR yet

`text` submodule in GHC source tree already points to https://gitlab.haskell.org/ghc/packages/text/-/tree/wip/windows-clang, so you likely need it in `text` master branch to benefit from your work here.

TerrorJack:
Yes, my branch was originally on top of what GHC source tree points to, therefore including that patch. I believe @bgamari is the right one to open a PR for that one :)

Bodigrim:
Up to you, whoever ends up raising a PR for 5a43fecb62e9769c4cf8b456c7b273901048ea72, please coordinate with @hasufell to publish an unofficial bindist and add a CI job.

text.cabal
```diff
@@ -94,6 +94,12 @@ library
       extra-libraries: c++
     elif os(openbsd)
       extra-libraries: c++ c++abi pthread
+    elif os(windows) && impl(ghc > 9.3)
+      extra-libraries: c++ c++abi
+    elif arch(wasm32)
+      cpp-options: -DSIMDUTF_NO_THREADS
+      cxx-options: -fno-exceptions
+      extra-libraries: c++ c++abi
```
Bodigrim, line 102: You can workaround `arch(wasm32)` by swapping branches:
```cabal
    elif arch(linux)
      extra-libraries: stdc++
    else 
      cpp-options: -DSIMDUTF_NO_THREADS
      cxx-options: -fno-exceptions
      extra-libraries: c++ c++abi
```
Actually this is no worse than before: it's not like `extra-libraries: stdc++` is correct for, say, `arch(netbsd)` or other exotic systems.

Issue #304, opened 2020-11-10
Add fromPtr variant that is NULL-terminated


sjshuck:
For my case, I need to copy a region of UTF-16 text from the heap into a `Text`.  It is NULL (U+0000) terminated.

Currently I'm scanning each 16-bit code unit in a loop to detemine the length and then calling `Text.fromPtr`, but I wonder if an easier method can be provided.

Bodigrim:
@sjshuck isn't [cstringLength#](https://hackage.haskell.org/package/base-4.16.1.0/docs/GHC-Exts.html#v:cstringLength-35-) enough for your purposes?

sjshuck:
Sure.  I didn't look at how it's implemented; the docs don't discuss; I'd expect it to be something like `strlen()`.  In any case I'd like it if `text` exposed a higher-level function that didn't require me to dip into `GHC.Exts` and `MagicHash`.

Asking directly: Would you be opposed to
```haskell
fromPtr0 :: Ptr Word8 -> IO Text
fromPtr0 ptr = do
    len <- c_strlen $ castPtr ptr
    fromPtr ptr (fromIntegral len)
```
?

Bodigrim:
Looks good to me, feel free to raise a PR.

PR #423, opened 2022-04-15
Add fromPtr0

sjshuck:
Fixes haskell/text#304.


sjshuck:
Most of the noise is moving `addrLen` to a more general module for reuse.

src/Data/Text/Internal/Private.hs
```diff
@@ -11,7 +11,8 @@
 
 module Data.Text.Internal.Private
     (
-      runText
+      addrLen
```
Bodigrim, line 14: Just export `addrLen` from `Data.Text.Show` and import it in `Data.Text.Foreign` please.

sjshuck: You got it.

Bodigrim: I meant to revert `Data.Text.Internal.Private` back: do not move `addrLen` outside of `Data.Text.Show`, just export it from there directly.

sjshuck: Yep, I understood.

Bodigrim (approved):

Bodigrim:
Thanks @sjshuck!

PR #429, opened 2022-04-25
Don't link against libgcc_s on Windows with GHC 9.4 and later

bgamari:
As of GHC 9.4 we use libc+++ and Clang on Windows.

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -94,6 +94,8 @@ library
       extra-libraries: c++
     elif os(openbsd)
       extra-libraries: c++ c++abi pthread
+    elif os(windows) && impl(ghc > 9.3)
+      extra-libraries: c++ c++abi
     else
       extra-libraries: stdc++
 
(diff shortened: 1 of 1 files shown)


Bodigrim:
@bgamari 

1. Please use the same GHC version in both `impl()`. Currently the first one is `impl(ghc < 9.4)` and another one is `impl(ghc < 9.3)`.
2. The proposed patch enables `cpp-options: -DSIMDUTF_NO_THREADS` and `cxx-options: -fno-exceptions` for Windows + GHC 9.4. Is it intended?

bgamari:
> Please use the same GHC version in both `impl()`. Currently the first one is `impl(ghc < 9.4)` and another one is `impl(ghc < 9.3)`.

Sigh, quite right. I'll fix.

> The proposed patch enables `cpp-options: -DSIMDUTF_NO_THREADS` and `cxx-options: -fno-exceptions` for Windows + GHC 9.4. Is it intended?

Hmm, yes, I didn't read that comment closely enough. Indeed we likely shouldn't do this on Windows.

Bodigrim (approved):

Bodigrim:
@bgamari thanks, please rebase.

Bodigrim:
Nevermind, I can squash on my end. Thanks!

PR #431, opened 2022-05-11
Add withCString

dpwiz:
Like #254, but only adds one function. `peekCString` is basically `fromPtr0`.

### DIFF
--- a/src/Data/Text/Foreign.hs
+++ b/src/Data/Text/Foreign.hs
@@ -21,6 +21,7 @@ module Data.Text.Foreign
     , useAsPtr
     , asForeignPtr
     -- ** Encoding as UTF-8
+    , withCString
     , peekCStringLen
     , withCStringLen
     -- * Unsafe conversion code
@@ -40,10 +41,11 @@ import Data.Text.Internal.Unsafe (unsafeWithForeignPtr)
 import Data.Text.Show (addrLen)
 import Data.Text.Unsafe (lengthWord8)
 import Data.Word (Word8)
-import Foreign.C.String (CStringLen)
+import Foreign.C.String (CString, CStringLen)
 import Foreign.ForeignPtr (ForeignPtr, mallocForeignPtrArray)
 import Foreign.Marshal.Alloc (allocaBytes)
 import Foreign.Ptr (Ptr, castPtr)
+import Foreign.Storable (pokeByteOff)
 import GHC.Exts (Ptr(..))
 import qualified Data.Text.Array as A
 
@@ -148,6 +150,21 @@ asForeignPtr t@(Text _arr _off len) = do
   fp <- mallocForeignPtrArray len
   unsafeWithForeignPtr fp $ unsafeCopyToPtr t
   return (fp, I8 len)
+
+-- | Marshal a 'Text' into a C string with a trailing NUL byte,
+-- encoded as UTF-8 in temporary storage.
+--
+-- The temporary storage is freed when the subcomputation terminates
+-- (either normally or via an exception), so the pointer to the
+-- temporary storage must /not/ be used after this function returns.
+--
+-- @since 2.0.1
+withCString :: Text -> (CString -> IO a) -> IO a
+withCString t@(Text _arr _off len) action =
+  allocaBytes (len + 1) $ \buf -> do
+    unsafeCopyToPtr t buf
+    pokeByteOff buf len (0 :: Word8)
+    action (castPtr buf)
 
 -- | /O(n)/ Decode a C string with explicit length, which is assumed
 -- to have been encoded as UTF-8. If decoding fails, a


src/Data/Text/Foreign.hs
```diff
@@ -149,6 +151,23 @@ asForeignPtr t@(Text _arr _off len) = do
...
 
+-- | Marshal a 'Text' into a C string with a trailing NUL byte,
+-- encoded as UTF-8 in temporary storage.
+--
+-- The temporary storage is freed when the subcomputation terminates
+-- (either normally or via an exception), so the pointer to the
+-- temporary storage must /not/ be used after this function returns.
+--
+-- @since 2.3
+withCString :: Text -> (CString -> I8 -> IO a) -> IO a
```
Bodigrim, line 162: Any pointers to functions with a similar design in other packages?

src/Data/Text/Foreign.hs
```diff
@@ -149,6 +151,23 @@ asForeignPtr t@(Text _arr _off len) = do
...
   return (fp, I8 len)
 
+-- | Marshal a 'Text' into a C string with a trailing NUL byte,
+-- encoded as UTF-8 in temporary storage.
+--
+-- The temporary storage is freed when the subcomputation terminates
+-- (either normally or via an exception), so the pointer to the
+-- temporary storage must /not/ be used after this function returns.
+--
+-- @since 2.3
```
Bodigrim, line 161:
```suggestion
-- @since 2.0.1
```

src/Data/Text/Foreign.hs
phadej: `I8` doesn't look like a right type for the length.

`bytestring` has https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Char8.html#v:useAsCStringLen, with ` (CStringLen -> IO r) -> IO r` type. That however doesn't add a zero.

`base` has https://hackage.haskell.org/package/base-4.16.1.0/docs/Foreign-C-String.html#v:withCStringLen.
Its `withCString` adds a NUL byte, and `withCStringLen` doesn't. Interesting choice.

Having a variant with length and trailing zero make sense, and the buffer is copied anyway, so it's cheap to just add zero.

phadej: EDIT: yeah, `text` actually has `withCStringLen` https://hackage.haskell.org/package/text-2.0/docs/Data-Text-Foreign.html#v:withCStringLen which doesn't add a null, and **doesn't** have `withCString` which adds a null and doesn't give you a length. But IMHO it makes sense to have all three variants:

1. `withCString :: Text -> (CString -> IO r) -> IO r` null-terminated
2. `withCStringLen :: Text -> (CStringLen -> IO r) -> IO r` with length, but not null terminated
3. `withCStringLenAndNul :: Text -> (CStringLen -> IO r) -> IO r` with length but also null terminated.

dpwiz: IMO using `CStringLen` for null-terminated c-strings is confusing. 

GHC uses this type distinction to mark 0-terminated strings differently from explicitly sized arrays:
https://hackage.haskell.org/package/base-4.16.1.0/docs/Foreign-C-String.html#v:withCString

dpwiz: > I8 doesn't look like a right type for the length.

Ugh.. I thought it was made specifically to mark length-in-bytes to prevent confusing it with length-in-whatever-the-user-thinks-it-is. But now I see it has no constructor published, forcing users to jump through the hoops.

phadej: @dpwiz different type?

```haskell
type CString = Ptr CChar
type CStringLen = (Ptr CChar, Int)
```

They are just aliases to `Ptr CChar`.

dpwiz: Different names, whatever. The point being, there is a convention that `CString` is *used* in different places than `CStringLen`. Adding things like `withCStringLen0` disrupts this convention.
According to it `withCStringLenAndNul` should be `withCStringAndLen`.

dpwiz: A situation I'm worried about goes like this:

Some imaginary lib uses the following functions:

- `Widgets.useNulTerminatedUtf8 :: CString -> IO ()`
- `Widgets.useSizedUtf8 :: CStringLen -> IO ()`

Now, what would be a correct pick?

- `Text.withCString :: Text -> (CString -> IO ()) -> IO ()` -- obvious, `useNulTerminatedUtf8`
- `Text.withCStringLen :: Text -> (CStringLen -> IO ()) -> IO ()` -- obvious, `useSizedUtf8`
- `Text.withCStringLenAndNul :: Text -> (CStringLen -> IO ()) -> IO ()` -- who to believe here, the function name or the type name? Ignore both and actually read the docs.

dpwiz:
So, what's would be the next step in this?

I can remove the I8 argument, bringing the function in line with GHC `withCString`. Users can get text length for free with `(+ 1) . lengthWord8`.
`withCStringLenAndNul` can be judged separately later.

src/Data/Text/Foreign.hs
phadej: @dpwiz but this, your PR adds `??? :: Text -> (CString -> Int -> IO r) -> IO r` (length + null) variant, with a name of `Text -> (CString -> IO r) -> IO r` function (if we are to follow naming conventions of other libraries).

Please amend the PR to what you actually want. Currently either the name or the type are wrong.

dpwiz: Done

Lysxia:
> Users can get text length for free with `(+ 1) . lengthWord8`.

I think that's worth mentioning in the doc.

Otherwise, LGTM. For reference, this is essentially the same logic as `Data.ByteString.useAsCString`.

Bodigrim (approved):

Mikolaj:
So is `Data.Text.Foreign.useAsCString` planned or is `withCString` close enough (I haven't yet wrapped my head around it, but I need to decide whether to wait for text 2.1 with my update of sdl2-ttf to text2 or just use `withCString`)?

Bodigrim:
@Mikolaj `withCString` is `useAsCString` in disguise.

Mikolaj:
Indeed, it's precisely what I needed and it works fine: https://github.com/haskell-game/sdl2-ttf/pull/16

PR #436, opened 2022-05-16
Add foldr' and document the laziness of foldr

Lysxia:
Unlike for lists, `foldr'` is a useful primitive for strict `Text` since we can easily iterate from either end. I discovered that while working on the other PR about `statefulSpan`, which uses `foldr'` in a test. It also turns out that `foldr` is lazier than one might expect, thanks to its implementation via a lazy stream, rather than an eager imperative loop, so that seemed worth documenting.

`foldr` and `foldr'` are thus mostly equivalent *denotationally*, ignoring the strictness differences, but *operationally*, `foldr` starts executing from the left, while `foldr'` starts from the right. `foldr` can short-circuit, whereas `foldr'` has to traverse the whole list, but can do so in a tighter loop.

- `foldr'` can also be defined for lazy `Text` but it's a little more involved.
- Dually, `foldl` can also be made lazier than it is currently, by starting from the right end (and this is arguably evidence that `foldr`'s aforementioned laziness is not that obvious from a certain perspective), but because that's an existing function I will need a more careful look to be sure I'm not breaking anything. Hence I'm postponing that for a future PR.

Note: this idea is already present in bytestring https://github.com/haskell/bytestring/blob/master/Data/ByteString.hs#L558

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -98,6 +98,7 @@ module Data.Text
     , foldl1
     , foldl1'
     , foldr
+    , foldr'
     , foldr1
 
     -- ** Special folds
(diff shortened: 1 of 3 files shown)


Bodigrim (commented):
Would it be possible to add a test, which witnesses the difference between `foldr` and `foldr'`? E. g., use `f` which explodes with an `error` on certain chars and a carefully crafted `Text`.

Lysxia:
I added a test to distinguish `foldr` from `foldr'` and also `foldl`/`foldl'`.

I could also test for the memory usage of `foldr'`, with a basic example `foldr' (+) 0` on a long text with a restricted stack/heap. I have a working prototype but I'm not sure of the best way to include it in the repo. Is it worth a separate test suite with its own RTS options, or is there a better alternative?

Bodigrim (approved):

Issue #362, opened 2021-08-11
spanWith, takeWhileWith


devvythelopper:
It would be nice to have a variant of `span` and a variant of `takeWhile` that do not just apply a predicate, but also fold over a value.

The practical application I have in mind is to `span` or `takeWhile` but only up to a maximum number of characters. This of course could be implemented if there was a `spanMax` and `takeWhileMax` but a generalized `spanWith` and `takeWhileWith` that also do a fold (in this case over the number of characters already selected by the predicate) seems to be the better choice to me.

lehmacdj:
`takeWhileMax n predicate = takeWhile predicate . take n` and should have fairly similar performance characteristics to the dedicated version due to fusion so at least for your practical application I think the current interface is fine. 

The more general version maybe still is necessary if you want to do something more complicated though; I don't see a way to cobble it together out of existing library functions in a good way.

phadej:
@lehmacdj there is no fusion (#348)

lehmacdj:
Good to know, sad to hear though, but the discussion in that linked issue makes sense

Lysxia:
This is implemented in PR #437 (which was originally motivated by issue #178, which is the same issue as this).

PR #437, opened 2022-05-16
Stateful span

### LINKED ISSUE #178: Add stateful 'span'?
O(n) indexing makes some algorithms hard to express. For instance, let's say that I want to break a string into words, but avoid breaking parenthesized groups: `"(a b) c [d e]"` becomes `["(a b)", "c", "[d e]"]`. At the moment, this doesn't seem to be easy to write without incurring a 2× performance penalty (though I might be mistaken). So, I propose adding a function along these lines:

```haskell
-- name subject to bikeshedding
statefulSpan :: (a -> Char -> (a, Bool)) -> a -> Text -> (a, Text, Text)
```

Lysxia:
Closes #178 and #362  
Depends on PR #436

This PR implements `spanM` (span from the left) and `spanEndM`  (span from the right) for both strict and lazy `Text`.

`spanM` is a combination of `span` and `foldr`.

- `spanM`, as its name indicates, generalizes `span`, with a monadic predicate, of type `Char -> m Bool` rather than `Char -> Bool`.
- `spanM` generalizes a short-circuiting `foldr`, by also remembering the location where it stopped so you can easily resume with another search from there.
- `spanM` can also implement `foldl'` with the same space and time efficiency, by making the predicate always `True` and only using the state effect.

I argue these two functions pass the Fairbairn threshold as they provide the ability to safely move back and forth inside a `Text`. As mentioned in #178, this is currently not possible to do both efficiently and safely: either you keep track of an offset in code points, which then requires a redundant traversal to do `splitAt`, or you keep track of an offset in bytes, which is unsafe.

With `spanM` one can easily implement a `takeWhileM` (#362) by ignoring the remaining suffix. I'm not sure if its allocation would be optimized out already, but if not, I can generalize `spanM`` with some continuation-passing trick to do that in a future PR.

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -146,6 +146,8 @@ module Data.Text
     , breakOnEnd
     , break
     , span
+    , spanM
+    , spanEndM
     , group
     , groupBy
     , inits
@@ -206,7 +208,7 @@ module Data.Text
 
 import Prelude (Char, Bool(..), Int, Maybe(..), String,
                 Eq, (==), (/=), Ord(..), Ordering(..), (++),
-                Read(..),
+                Monad(..), pure, Read(..),
                 (&&), (||), (+), (-), (.), ($), ($!), (>>),
                 not, return, otherwise, quot, IO)
 import Control.DeepSeq (NFData(rnf))
@@ -258,6 +260,8 @@ import System.Posix.Types (CSsize(..))
 import System.IO.Unsafe (unsafePerformIO)
 
 -- $setup
+-- >>> :set -package transformers
+-- >>> import Control.Monad.Trans.State
 -- >>> import Data.Text
 -- >>> import qualified Data.Text as T
 -- >>> :seti -XOverloadedStrings
@@ -1428,7 +1432,7 @@ splitAt n t@(Text arr off len)
 -- | /O(n)/ 'span', applied to a predicate @p@ and text @t@, returns
 -- a pair whose first element is the longest prefix (possibly empty)
 -- of @t@ of elements that satisfy @p@, and whose second is the
--- remainder of the list.
+-- remainder of the text.
 --
 -- >>> T.span (=='0') "000AB"
 -- ("000","AB")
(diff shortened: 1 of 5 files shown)


Bodigrim:
How about `spanM :: Monad m => (Char -> m Bool) -> Text -> m (Text, Text)`? It is more general and easier to guess semantics from the type signature.

Lysxia:
> How about `spanM :: Monad m => (Char -> m Bool) -> Text -> m (Text, Text)`?

Very nice idea, thanks!

src/Data/Text.hs
```diff
@@ -1422,6 +1450,51 @@ break :: (Char -> Bool) -> Text -> (Text, Text)
...
+-- 'span' is 'spanM' specialized to 'Data.Functor.Identity.Identity':
+--
+-- @
+-- -- for all p :: Char -> Bool
+-- 'span' p = 'Data.Functor.Identity.runIdentity' . 'spanM' ('pure' . p)
+-- @
+spanM :: Monad m => (Char -> m Bool) -> Text -> m (Text, Text)
+spanM p t@(Text arr off len) = go 0
+  where
+    go !i | i < len = case iter t i of
```
Bodigrim, line 1469: It's (marginally) faster to use `iterArray`.

tests/Tests/Properties/Substrings.hs
```diff
@@ -157,6 +159,48 @@ t_inits           = L.inits       `eqP` (map unpackS . T.inits)
 tl_inits          = L.inits       `eqP` (map unpackS . TL.inits)
 t_tails           = L.tails       `eqP` (map unpackS . T.tails)
 tl_tails          = L.tails       `eqPSqrt` (map unpackS . TL.tails)
+
+-- To generate function arguments for spanM:
+-- - we fix the accumulator to an Int counter, without loss of generality
+-- - we reduce the probability of producing False, by first generating an Int
+--   (here in [-5, 5]) and then mapping one outcome (0) to False and the others to True.
+genSpanFun :: Gen (Fun (Int, Char) Bool)
+genSpanFun = (fmap . fmap) (/= (0 :: Int)) (resize 5 arbitrary)
```
Bodigrim, line 168: I'd rather define 

```haskell
newtype SkewedBool = Skewed Bool 

instance Arbitrary SkewedBool where 
  arbitrary = frequency [(1, pure $ Skewed False), (5, pure $ Skewed True)]
```

Then `Fun (Int, Char) SkewedBool` automatically has desired distribution and you do not need boilerplate around `forAllShrink` below.

src/Data/Text/Lazy.hs
```diff
@@ -1286,14 +1301,57 @@ break p t0 = break' t0
...
+-- -- for all p :: Char -> Bool
+-- 'span' p = 'Data.Functor.Identity.runIdentity' . 'spanM' ('pure' . p)
+-- @
+spanM :: Monad m => (Char -> m Bool) -> Text -> m (Text, Text)
+spanM p t0 = go t0
+  where
+    go Empty = pure (empty, empty)
+    go (Chunk t ts) = do
+        (t1, t2) <- T.spanM p t
+        if T.null t2 then (\(t3, t4) -> (chunk t1 t3, t4)) <$> go ts
```
Bodigrim, line 1331:
```suggestion
        if T.null t2 then first (chunk t) <$> go ts
```

Bodigrim (commented):
Please rebase.

Bodigrim (approved):

PR #444, opened 2022-06-21
Emphasize total functions in Data.Text.Encoding

### LINKED ISSUE #248: Improve Data.Text.Encoding docs
Total functions should come before their partial variants in the Haddocks, and partial functions should be labeled as such.

(This doesn't matter to experienced Haskellers, but is a stumbling point for beginners, especially since they already have trouble with string handling).

tbidne:
Resolves #248. The basic idea is to split total/partial functions into explicit sections, and place the total section first. Some comments/questions:

1. I am open to making the warnings on partial functions themselves louder e.g. changing `decodeUtf8`'s current text:

    > This is a partial function: it checks that input is a well-formed
     UTF-8 sequence and copies buffer or throws an error otherwise.

    to (up to bikeshedding):

    >  __WARNING__: This is a __partial__ function: it checks that input is a well-formed
     UTF-8 sequence and copies buffer or throws an error otherwise.
 
    I don't think the danger can really be overemphasized, given how destructive they have proven to be. And this way people using e.g. HLS can see them on hover even if they are not looking at the module's documentation. But I wanted to solicit feedback before becoming too opinionated.

1. Minor point, but is there a way to link anchors without the qualified module name showing up (e.g. either a 'local' anchor or supplying a custom name? That is, I linked to the total section with `See "Data.Text.Encoding#g:total" ...`, and this renders as:

    > See [Data.Text.Encoding](https://tbidne.github.io/text/Data-Text-Encoding.html#g:total) ...
    
    This is slightly annoying imo as the link text is not indicative of the actual location (specifically, the `total` anchor), though it does work, at least. I suppose one could supply a full hyperlink, but may this isn't a big deal in any case. [Here](https://haskell-haddock.readthedocsbe.io/en/latest/markup.html#anchors) are the haddock docs for reference.

1. Presumably we would want the same thing done to `Data.Text.Lazy.Encoding`. I will do this after the language is tentatively approved.

1. I noticed in the unmerged https://github.com/haskell/text/pull/318, there is a note about updating the changelog. I can do this after everything is settled.

The rendering can be seen [here](https://tbidne.github.io/text/Data-Text-Encoding.html).

Thanks!

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -25,30 +25,38 @@ module Data.Text.Encoding
     (
     -- * Decoding ByteStrings to Text
     -- $strict
-      decodeASCII
-    , decodeLatin1
+
+    -- ** Total Functions #total#
+    -- $total
+      decodeLatin1
+    , decodeUtf8Lenient
+
+    -- *** Catchable failure
+    , decodeUtf8'
+
+    -- *** Controllable error handling
+    , decodeUtf8With
+    , decodeUtf16LEWith
+    , decodeUtf16BEWith
+    , decodeUtf32LEWith
+    , decodeUtf32BEWith
+
+    -- *** Stream oriented decoding
+    -- $stream
+    , streamDecodeUtf8With
+    , Decoding(..)
+
+    -- ** Partial Functions
+    -- $partial
+    , decodeASCII
     , decodeUtf8
     , decodeUtf16LE
     , decodeUtf16BE
     , decodeUtf32LE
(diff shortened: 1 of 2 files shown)


Bodigrim (approved):
LGTM. 
Shall we move `decodeUtf8Lenient` out of "Controllable error handling" group?

tbidne:
> Shall we move `decodeUtf8Lenient` out of "Controllable error handling" group?

I agree. The top-level `total` section seems more appropriate.

Bodigrim:
Thanks!

PR #447, opened 2022-06-24
Document that unpack . pack is not id

Bodigrim:


src/Data/Text.hs
```diff
@@ -443,7 +454,10 @@ compareText (Text arrA offA lenA) (Text arrB offB lenB) =
 -- * Conversion to/from 'Text'
 
 -- | /O(n)/ Convert a 'String' into a 'Text'.
--- Performs replacement on invalid scalar values.
+-- Performs replacement on invalid scalar values, so 'unpack' . 'pack' is not 'id':
```
Lysxia, line 457:
```suggestion
-- Performs replacement on invalid scalar values, so @'unpack' . 'pack'@ is not 'id':
```

Bodigrim:
@Lysxia  done.

Lysxia (approved):

PR #448, opened 2022-06-27
Allow caller more control in stream decoding

david-sledge:
I've been wanting this feature for a while, so I decided to prototype it and create a pull request.


Lysxia:
Rather than STT, safer alternatives would be to specialize this to one specific monad (was state the original motivation?) or to use a different algorithm that does not involve STT.

Lysxia:
STT is notably unsafe with continuations.

Bodigrim:
`text` is a boot library, it cannot depend on `STMonadTrans` package, unfortunately. 

I'm not very happy with the API of `Data.Text.Encoding`. The fact that the only way to abort decoding is to raise an `error` (and `catch` it elsewhere) is a joke. Ideally it should be possible to drive decoding from without, e. g., 

```haskell
decodeUtf8With2 :: ByteString -> ByteString -> Either Int (Text, ByteString)
```

Then clients are free to interpret `Int` (the position, where decoding failed) as they like, in whatever monad.

david-sledge:
> `text` is a boot library, it cannot depend on `STMonadTrans` package, unfortunately.

Removed.

> I'm not very happy with the API of `Data.Text.Encoding`. The fact that the only way to abort decoding is to raise an `error` (and `catch` it elsewhere) is a joke. Ideally it should be possible to drive decoding from without, e. g.,
> 
> ```haskell
> decodeUtf8With2 :: ByteString -> ByteString -> Either Int (Text, ByteString)
> ```

Done. Check out the test case `t_decode_withM_error5` in tests/Tests/Properties/Transcoding.hs. There other examples, too, for Cont and Maybe. What's more is that the error handlers of the additional functions (the ones ending with `M`) take the byte position where the error occurred.

Lysxia:
This still has the same unsafety as STT, since it's just an inlining of the STT logic. You can cause a segfault by specializing to `ContT` and by using a continuation twice.
Taking a callback is fundamentally flawed, so that the current implementation cannot simply be generalized to be in an arbitrary monad. Changing the API as bodigrim suggested seems like a better option (you may also remember the part that was decoded so far).

david-sledge:
Ahhh... I see what you mean, and it'd be a problem with the list monad, too. However, I noticed that `streamDecodeUtf8With` returns a continuation in the `Decoding` data type. I looked how the risk of calling that continuation multiple times is mitigated. Essentially, the continuation a wholly separate invocation of a function. It calls `runST` in which the text array is created, populated, frozen, embedded in the Text data type, and returned without being modified any further.

So I applied the same concept around the call to the error handler. Before the error handler is called, it runs through the processes just before exiting; the text array is frozen, no longer modified, and the reference to last state value is discarded. Then after the error handler returns, it simulates `runST` with a call to `runRW#`, makes a copy of the text array where it left off, and continues its operations on the copy.

Lysxia:
Changing the API to not take a callback at all and instead return an `Either` that tells you exactly where it stopped would still be much simpler, safer, and faster. I would prefer to consider that first.

Passing the `State#` token explicitly like this, while in an arbitrary monad, makes this quite difficult to review with confidence.

david-sledge:
While I understand the desire to use `Either`, that doesn't really get me closer to my goal, which is why I was going for arbitrary monad which would meet my goal and allow for `Either`. Since insert-your-favorite-monads-here option has been removed from the table, I found myself at an impasse. So I took a step back, looked at the issue again, and reevaluated the problem.

With the requirements and constraints of
* Allow a means to abort decoding without raising an `error`.
* No STT.
* No arbitrary monad implementation.
* No manually passing the `State#` token.
* Indicate position in source `ByteStream` where the error occurs.

I think I found a solution that'll fit the bill.

Lysxia:
This is actually close to what we were suggesting, and indeed much more satisfactory.

david-sledge:
Excellent! A couple of questions:
* The byte position does not currently track across continuation invocations and instead resets back to zero. What's the opinion on whether it should track or the caller be responsible for tracking the cumulative byte position/total bytes read?
* What version should the `@since` annotation be for `streamDecodeUtf8With'`? 2.0.1, 2.1.0, or do I not need to worry about it yet?

david-sledge:
For the first question, I went ahead and let the library track the absolute position for the caller.

david-sledge:
Stream decoders added for utf-16 and utf-32, and an `Either` decoder for ASCII. I'll work on the lazy options next.

src/Data/Text/Encoding.hs
```diff
@@ -213,6 +218,20 @@ isValidBS bs = start 0
...
 #endif
 
+-- | Decode a 'ByteString' containing 7-bit ASCII encoded text.
+--
+-- This is a total function: On success the decoded 'Text' is within a
+-- 'Right' value, and an error ('Left' 'Int') indicates the postion of
+-- the offending 'Word8'.
+decodeAsciiE :: ByteString -> Either Int Text
+decodeAsciiE bs = withBS bs $ \fp len -> if len == 0 then Right empty else runST $ do
+  asciiPrefixLen <- fmap cSizeToInt $ unsafeIOToST $ unsafeWithForeignPtr fp $ \src ->
```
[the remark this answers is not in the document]
david-sledge: It's a direct copy and paste of `decodeASCII` with the bare minimum changes so that `Either` is used instead of calling `error`.

src/Data/Text/Encoding.hs
```diff
@@ -406,11 +344,199 @@ streamDecodeUtf8With ::
...
+  -- the position in the 'ByteString' where the code point starts,
+  -- the incomplete code point, and a function that accepts another
+  -- 'ByteString' as a continuation of the previous input.
+  = ThusFar !Text !Int !ByteString (ByteString -> StreamDecode w)
+  -- | An encounter witt an invalid utf-8 'Word8'. The value contains
+  -- the decoded text up to but not including the invalid 'Word8',
+  -- the position in the 'ByteString' of the offending 'Word8', the
+  -- offender, and a function that accepts a possible 'Char' as to
+  -- interpret the 'Word8' in this specific occurrence of it.
+  | InvalidWord !Text !Int !w (Maybe Char -> StreamDecode w)
```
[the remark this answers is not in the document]
david-sledge: I deliberately left out the third constructor, because I found it redundant with `Incomplete` when the remainder `ByteString` is empty. Also the continuation tracks the byte position across multiple continuation calls for convenience. Just because the `ByteString` was decoded successfully and completely for a given iteration, doesn't mean the caller has finished feeding it data.

However, the responsibility tracking total `Word8`s processed could be pushed to the caller. So if we're discussing simplifying the return type, I would go with something like:
   ```
   -- | Result of decoding
   data StreamDecode w = StreamDecode
     !(Maybe w)  -- ^ If an invalid code point was encountered
     !Text       -- ^ Longest valid prefix
     !ByteString -- ^ Remaining undecoded data after the possible invalid code point
     !Int        -- ^ Byte position of remaining undecoded data
   ```
So complete success with a given `ByteString` input would be represented with `Nothing` for the invalid code point and an empty `ByteString` suffix.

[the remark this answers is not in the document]
david-sledge: So I wrote decoders without the recursion, and using those I wrote decoders with the recursions. Looking at the two, there doesn't seem to be any benefit to non-recursive over the recursive. There's nothing a caller can do with former that can't be done with the latter. If the caller chooses to ignore the continuation function, it's effectively the same as the non-recursive. Also, as I was writing the test cases, the recursion turned out to be a little more convenient.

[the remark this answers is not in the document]
Lysxia: Thanks for doing this. The way I would compare the two instead is that using the non-recursive one you can still easily implement the recursive one, whereas if you try the other way around, it will allocate an unneeded continuation which won't be optimized out because recursive functions are not inlined.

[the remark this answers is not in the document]
david-sledge: I removed the `decodeUtfXStream` functions, but kept the `chunksDecoderToStream` function and moved it to `Data.Text.Encoding.Types`. So a streaming option still remains available, and is simple to produce. Also, since `DecodeResult` contains the remaining unencoded `ByteString`, I make the chunk decoders take two `ByteString` arguments so the caller doesn't have to append the leftovers of the previous call to the new `ByteString` of the next call.

changelog.md
```diff
@@ -1,3 +1,16 @@
+### 2.0.0.1
```
Bodigrim, line 1:
```suggestion
### 2.0.1
```

src/Data/Text/Internal/Builder.hs
```diff
@@ -181,6 +186,12 @@ fromText t@(Text arr off l)
 --
 --  * @'toLazyText' ('fromString' s) = 'L.fromChunks' [S.pack s]@
 --
+-- Performs replacement on invalid scalar values:
+--
+-- >>> fromString "\55555"
+-- "\65533"
+--
+-- @since 1.2.0.0
```
Bodigrim, line 194: Are you sure? It seems available earlier.

tests/Tests/Properties/Transcoding.hs
```diff
@@ -1,13 +1,17 @@
...
-{-# LANGUAGE OverloadedStrings, ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings, ScopedTypeVariables, FlexibleContexts #-}
 {-# OPTIONS_GHC -fno-warn-missing-signatures #-}
 module Tests.Properties.Transcoding
     ( testTranscoding
     ) where
 
+import Control.Monad.Writer (runWriterT, tell)
+import Data.Functor.Identity (runIdentity)
+-- import Data.Monoid (Endo(appEndo))
```
Bodigrim, line 11: Please remove the commented line.

text.cabal
```diff
@@ -255,6 +256,7 @@ test-suite tests
     deepseq,
     directory,
     ghc-prim,
+    mtl,
```
Bodigrim, line 259: Is it possible to avoid dependency on `mtl`?

src/Data/Text/Internal/Builder.hs
david-sledge: That's inadvertent from a copy-paste rollback. It'll be removed.

tests/Tests/Properties/Transcoding.hs
david-sledge: Will do.

text.cabal
david-sledge: Yes

Bodigrim:
We are getting closer to the final cut off before GHC 9.4. @Lysxia could you please complete your review on this? I don't have much bandwidth at the moment.

Bodigrim:
I'm not sure exactly. Given that @bgamari has not chases us yet, it would probably take at least a week or more.

src/Data/Text/Encoding.hs
```diff
@@ -267,54 +349,112 @@ decodeUtf8With2 onErr bs1@(B.length -> len1) bs2@(B.length -> len2) = runST $ do
...
+--
+-- @since 2.0.1
+decodeUtf32Chunks
+  :: Bool       -- ^ Indicates whether the encoding is big-endian (`True`) or little-endian (`False`)
+  -> ByteString -- ^ The first `ByteString` chunk to decode. Typically this is the unencoded data from the previous call of this function.
+  -> ByteString -- ^ The second `ByteString` chunk to decode.
+  -> DecodeResult Text ByteString Word32
+decodeUtf32Chunks isBE = decodeChunksProxy noOptimization $ \ index _ srcOff ->
+  -- get next Word8 quartet
+  case (queryUtf32Bytes . index $ if isBE then srcOff else srcOff + 3)
```
Lysxia, line 416: Like above, it seems better to just inline `queryUtf32Bytes` here.

src/Data/Text/Encoding.hs
```diff
@@ -267,54 +349,112 @@ decodeUtf8With2 onErr bs1@(B.length -> len1) bs2@(B.length -> len2) = runST $ do
...
+      >>= ($ (index $ srcOff + (if isBE then 1 else 2)))
+      >>= ($ (index $ srcOff + (if isBE then 2 else 1))) of
+    Just f -> WriteAndAdvance (f . index $ if isBE then srcOff + 3 else srcOff) . const $ srcOff + 4
+    _ -> Invalid
+
+-- | Decode a 'ByteString' containing 7-bit ASCII encoded text.
+--
+-- This is a total function: On success the decoded 'Text' is within a
+-- 'Right' value, and an error ('Left' 'Int') indicates the postion of
```
Lysxia, line 425:
```suggestion
-- 'Right' value, and an error @('Left' 'Int')@ indicates the position of
```

src/Data/Text/Encoding.hs
```diff
@@ -213,45 +218,122 @@ isValidBS bs = start 0
...
-  outer marr len' 0 0
+data Progression
+  = WriteAndAdvance Char (Int -> Int)
+  | NeedMore
+  | Invalid
+
+decodeChunks :: (Bits w, Num w, Storable w) =>
+  w -- only used for Storable.sizeOf argument which the function discards
+  -> (ByteString -> ByteString -> Int -> Maybe (A.Array, Int))
+  -> ((Int -> Word8) -> Int -> Int -> Progression)
```
Lysxia, line 229: We can get rid of this continuation. Now that `decodeChunks` returns a `DecodeResult`, the caller can directly do the work of that continuation (including error recovery on invalid bytes), and then call `decodeChunks` again on the remaining chunk.

src/Data/Text/Encoding.hs
```diff
@@ -213,45 +218,122 @@ isValidBS bs = start 0
...
+  , srcOff < utf8End
+  -- potential valid utf8 content length
+  , utf8Len <- utf8End - srcOff
+  , bs' <- B.drop (srcOff - len1) $ B.take guessUtf8Boundary bs2
+  , isValidBS bs' = Just (runST $ do
+    marr <- A.new utf8Len
+    withBS bs' $ \fp _ -> unsafeIOToST $ unsafeWithForeignPtr fp $ \src ->
+      unsafeSTToIO $ A.copyFromPointer marr 0 src utf8Len
+    arr <- A.unsafeFreeze marr
+    pure $ (arr, utf8Len))
```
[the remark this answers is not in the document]
david-sledge: Indeed. That won't do.

src/Data/Text/Encoding.hs
```diff
@@ -267,54 +349,112 @@ decodeUtf8With2 onErr bs1@(B.length -> len1) bs2@(B.length -> len2) = runST $ do
...
+  -> ByteString -- ^ The first `ByteString` chunk to decode. Typically this is the unencoded data from the previous call of this function.
+  -> ByteString -- ^ The second `ByteString` chunk to decode.
+  -> DecodeResult Text ByteString Word16
+decodeUtf16Chunks isBE = decodeChunksProxy noOptimization $ \ index len srcOff ->
+  -- get next Word8 pair
+  let writeAndAdvance c n = WriteAndAdvance c $ const n
+      b0 = index $ if isBE then srcOff else srcOff + 1
+      b1 = index $ if isBE then srcOff + 1 else srcOff
+  in
+  case queryUtf16Bytes b0 of
```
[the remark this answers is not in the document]
david-sledge: Easily fixed.

src/Data/Text/Encoding.hs
```diff
@@ -213,45 +218,122 @@ isValidBS bs = start 0
...
+                  DecodeResult t Nothing bs' srcOff
+              invalid =
+                let srcOff' = srcOff + wordByteSize
+                    bytesToWord n word =
+                      if n > 0
+                        then bytesToWord (n - 1) $ (fromIntegral . index $ srcOff + wordByteSize - n) .|. (word `shiftL` 8)
+                        else word
+                in
+                contin srcOff' $ \ t bs' ->
+                  DecodeResult t (Just $ bytesToWord wordByteSize 0) bs' srcOff'
```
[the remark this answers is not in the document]
david-sledge: I thought about doing that and using a `Bool` instead of a `Maybe` to indicate a code point error. My hesitation is the type `DecodeResult !t !Bool !b !Int` implies that `Bool` could be `True` while `b` is empty. Admittedly, it can be ensured that such a thing is never the case, but it still feels wrong (for lack a better term).

src/Data/Text/Encoding.hs
david-sledge: That's not a continuation. That's the function that dictates by what rules to decode the `ByteStream` chunks into `Text`. `decodeChunks` is not exposed outside the module. It's just decoding processes common to all the chunk decoders. In fact, chunk decoders can be for ASCII, Latin-1, and CP-1252 just with providing the rules via the `decodeF` (and optionally `queryOptimization`).

Lysxia: Sorry I had a brainfart moment here!

src/Data/Text/Encoding.hs
```diff
@@ -213,45 +218,122 @@ isValidBS bs = start 0
...
-decodeUtf8With2 ::
-#if defined(ASSERTS)
-  HasCallStack =>
-#endif
-  OnDecodeError -> ByteString -> ByteString -> (Text, ByteString)
-decodeUtf8With2 onErr bs1@(B.length -> len1) bs2@(B.length -> len2) = runST $ do
-  marr <- A.new len'
-  outer marr len' 0 0
+data Progression
+  = WriteAndAdvance Char (Int -> Int)
```
Lysxia, line 222: Why is this a function rather than just an `Int` offset?

src/Data/Text/Encoding.hs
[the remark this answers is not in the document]
Lysxia: Isn't that already kind of an issue, in the case where the `Just w` is not actually the wrong word because it's actually the next one, one would have to go back and read from the bytestring to get the right diagnostic.

src/Data/Text/Encoding.hs
david-sledge: With UTF-8, the source offset determined by the number of words written to the destination array, so the argument is that number of words. This is not true for 16 and 32, so the number of words written is ignored.

src/Data/Text/Encoding.hs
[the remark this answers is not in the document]
david-sledge: You're right that it's not necessarily the problematic word of the code point, but it is the first word of the invalid code point (first non-decodable word?), even if the offending word follows two or three bytes later, and the documentation should reflect that. However, identifying the start of the code point was only one of purposes of `Just x`. The other purpose is that combined with rest of the unencoded `ByteString` it creates a non-empty `ByteString` type. Again, I wanted to avoid having the signature imply a possibly empty `ByteString` when there's an invalid code point. (I briefly considered using `DecodeResult Text (Either (w, ByteString) ByteString) Int` to emphasis this.)

src/Data/Text/Encoding.hs
Lysxia: In `decodeUtf8Chunks`, can't the source offset be deduced from `i` when `step` ends (which is currently discarded)?

Bodigrim:
Given https://github.com/haskell/text/pull/453#issuecomment-1191279987, I'll cut `text-2.0.1` from the current `master`, and this PR will be delayed until `text-2.0.2`. I'm happy to release `text-2.0.2` fairly soon, in time for GHC 9.4.2. Sorry @david-sledge, I hope it's not too much of delay for your purposes.

src/Data/Text/Encoding.hs
```diff
@@ -187,32 +189,295 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+  => w
+  -> ( Int
+    -> ByteString
+    -> A.MArray s
+    -> Int
+    -> Int
+    -> Int
+    -> ST s (Maybe (
+        (A.MArray s -> Int -> Int -> Int -> ST s (DecodeResult Text ByteString w))
+        -> ST s (DecodeResult Text ByteString w))
```
Lysxia, line 211: Isn't this just a Church-encoded tuple? Why not return a real tuple?

src/Data/Text/Encoding.hs
```diff
@@ -187,32 +189,295 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+      )
+    )
+  -> ((Int -> Word8) -> Int -> Int -> Progression)
+  -> ByteString
+  -> ByteString
+  -> ST s (DecodeResult Text ByteString w)
+decodeChunks w bulkDecodeF incrementalDecodeF bs1@(B.length -> len1) bs2@(B.length -> len2) = do
```
Lysxia, line 218: Is there no way to simplify this? as far as `decodeChunks` is concerned, `bulkDecodeF` parses a prefix of the source and copies it into the destination. This overlaps with the duties of `incrementalDecodeF`, which can be seen as parsing a one-character prefix. Can't there be a function that turns a `incrementalDecodeF` into a `bulkDecodeF`, so that `decodeChunks` then only takes a single function parameter?

src/Data/Text/Encoding.hs
```diff
@@ -187,32 +189,295 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+    outer dst dstLen = inner
+      where
+        inner srcOff dstOff
+          | srcOff < len1 = decode 0 bs1
+          | srcOff < len = decode len1 bs2
+          -- finished (for now)
+          | otherwise = goodSoFar
+          where
+            decode bsOff bs = do
+              mOuterArgs <- bulkDecodeF bsOff bs dst dstLen srcOff dstOff
```
Lysxia, line 243: Could this be O(n^(2)) in the case where `bs` is a bytestring which is almost UTF-8 except for an invalid byte at the end? I think what happens then is that `bulkDecodeF` scans the string, fails at the end, then `incrementalDecodeF` is called to advance a few bytes, and loop again, with `bulkDecodeF` being called at every iteration.

src/Data/Text/Encoding.hs
david-sledge: `Just outerArgs -> outerArgs outer` is less verbose than `Just (dst', dstLen', srcOff', dstOff') -> outer dst' dstLen' srcOff' dstOff'`, and at the time I had the idea that `bulkDecodeF` would be called at multiple places (which turned out not to be the case).

src/Data/Text/Encoding.hs
david-sledge: See my response to other concern you raised below.

src/Data/Text/Encoding.hs
david-sledge: Instead of `bulkDecodeF`, `decodeChunks` could accept a function that counts the number of continuous `Word8`s where the encodings overlap between UTF-8 and the encoding in question (e.g. `c_is_ascii`). Once it knows the length, it can copy that stretch of the bytestring, and then switch to `incrementalDecodeF` for the next code point. This also would require a different function for more than 8-bit word encodings, and simdutf might have to be abandoned.

Lysxia: That sounds like a good plan.

src/Data/Text/Encoding.hs
Lysxia: - It's less verbose but harder to understand.
- The type here is also not as precise as it should be, it contains more inhabitants than just Church encoded tuples, so it sends the reader on a goose chase to check that there are no additional tricks hiding there.
Maintenance-wise, a tuple will be less of a hassle.

david-sledge:
This branch no longer uses simdutf anymore. Should all of it and references to it be removed, or is it worth modifying simdutf to have a `determine_utf8_prefix_length` function?

Lysxia:
simdutf is there for performance, so if it is no longer used that needs to be justified by benchmarks. I suspect it's going to be difficult to compete with a vectorized implementation.

david-sledge:
I suspect with the issue you pointed out of where the next to last `Word8` is invalid, this reimplementation will perform better, because the current implementation of `decodeUtf8With2` has the same issue. However, in instances where the entire `ByteString` is valid, the current implementation will perform better. So one option is to only call `c_is_valid_utf8` once at the beginning of each `ByteString` evaluation. Another option is what I mentioned previously of adding to simdutf a function that counts the number of contiguous valid UTF-8 `Word8`s at the beginning of a `ByteString`.

Lysxia:
> So one option is to only call `c_is_valid_utf8` once at the beginning of each `ByteString` evaluation.

As I understand it this is what we would accept currently, we want the "valid" case to be as fast as possible (@Bodigrim is that right?), and slowing down in other cases is fine. Modifying the C++ function to also count how far the valid bytes extend so we don't have to loop in Haskell would be a nice bonus, but it's not a requirement.

Bodigrim:
> As I understand it this is what we would accept currently, we want the "valid" case to be as fast as possible (@Bodigrim is that right?), and slowing down in other cases is fine.

Correct.

> Another option is what I mentioned previously of adding to simdutf a function that counts the number of contiguous valid UTF-8 Word8s at the beginning of a ByteString.

That would be awesome, but I don't know how feasible it is. Maybe someone can check it with simdutf developers?

Lysxia (commented):
This is getting close

src/Data/Text/Encoding.hs
```diff
@@ -208,53 +214,120 @@ decodeAsciiPrefix bs = if B.null bs
...
+  then
+    if isValidBS 0 bs1Utf8Boundary bs1
+    then
+      if bs1Utf8Boundary < len1
+      then
+        let checkCodePointAccrossBoundary Reject _ = decodeResult True bs1Utf8Boundary
+            checkCodePointAccrossBoundary (Incomplete a) off = if off < len
+              then checkCodePointAccrossBoundary (utf8DetectContinue (index off) a) $ off + 1
+              else decodeResult False bs1Utf8Boundary
+            checkCodePointAccrossBoundary Accept off = wrapUpBs2 off
```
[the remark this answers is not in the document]
david-sledge: It's at the end of the first chunk and the beginning of the second. It validates a code point that crosses over the chunks. So this can happen, and when it does, the offset is one to three words into the second chunk.

david-sledge:
The error in the workflow seems to be due to an unrelated test. I've submitted an issue regarding it.

Bodigrim:
@david-sledge could you please rebase?

Bodigrim:
@Lysxia is it good to go?

Lysxia (changes requested):
Sorry this is taking so long @david-sledge. Tell me if you don't want to spend more time on this and I can take over.

1. In `DecodeResult`, the `Maybe w` is not worth the complexity of an extra type argument (which is also extra awkward to use). Even if you encode the invariant that the remaining suffix is not empty that way, the types don't enforce that the suffix is in fact invalid, so the calling code will still end up having to handle the vacuous case where it can't find a reason for the error.

2. Constructing the suffix by appending in the case where the error happens in the first chunk. Just return the index and let the caller decide what to do.

3. The handling of chunks gets convoluted in trying to deduplicate the logic between the two chunks. In practice, you're unlikely to just have to decode two arbitrary chunks. Ideally, you'd like to handle one chunk at a time, possibly with some residual state from a previous chunk, which contains just a single incomplete code point.

    In fact, the first chunk could rather be the decoder state at the end of the previous chunk, which also lets you throw the previous chunk away completely before getting to the next chunk.

4. Stylistic nit: it's easier to read `case` expressions with disjoint patterns (`Just _ | Nothing`) rather than wildcards (`Just _ | _`), because I don't have to look up previous clauses to find out what constructor is being handled.

david-sledge:
> Sorry this is taking so long @david-sledge. Tell me if you don't want to spend more time on this and I can take over.
> 
> 1. In `DecodeResult`, the `Maybe w` is not worth the complexity of an extra type argument (which is also extra awkward to use). Even if you encode the invariant that the remaining suffix is not empty that way, the types don't enforce that the suffix is in fact invalid, so the calling code will still end up having to handle the vacuous case where it can't find a reason for the error.
> 2. Constructing the suffix by appending in the case where the error happens in the first chunk. Just return the index and let the caller decide what to do.
> 3. The handling of chunks gets convoluted in trying to deduplicate the logic between the two chunks. In practice, you're unlikely to just have to decode two arbitrary chunks. Ideally, you'd like to handle one chunk at a time, possibly with some residual state from a previous chunk, which contains just a single incomplete code point.
>    In fact, the first chunk could rather be the decoder state at the end of the previous chunk, which also lets you throw the previous chunk away completely before getting to the next chunk.
> 4. Stylistic nit: it's easier to read `case` expressions with disjoint patterns (`Just _ | Nothing`) rather than wildcards (`Just _ | _`), because I don't have to look up previous clauses to find out what constructor is being handled.

Ok, I've a design that'll address your points including the quadratic time issue. It'll be a bit of a rewrite, so I'm going to convert this to a draft.

david-sledge:
> Sorry this is taking so long @david-sledge. Tell me if you don't want to spend more time on this and I can take over.

I'm actually rather enjoying this. Though I'm finding I have less time to work on it than I did before.
 
> 1. In `DecodeResult`, the `Maybe w` is not worth the complexity of an extra type argument (which is also extra awkward to use). Even if you encode the invariant that the remaining suffix is not empty that way, the types don't enforce that the suffix is in fact invalid, so the calling code will still end up having to handle the vacuous case where it can't find a reason for the error.

1. Removed.

> 2. Constructing the suffix by appending in the case where the error happens in the first chunk. Just return the index and let the caller decide what to do.

2. Errors are now returned as an `Int` which represents the start position of the erroneous code point.

> 3. The handling of chunks gets convoluted in trying to deduplicate the logic between the two chunks. In practice, you're unlikely to just have to decode two arbitrary chunks. Ideally, you'd like to handle one chunk at a time, possibly with some residual state from a previous chunk, which contains just a single incomplete code point.
>    In fact, the first chunk could rather be the decoder state at the end of the previous chunk, which also lets you throw the previous chunk away completely before getting to the next chunk.

3. Can't throw it completely away. The values themselves need to be preserved in case the next `ByteString` proves them to be valid. And can't copy it yet, because it could turn out to be in valid. However, I did change the logic so that the code point decoder state at the end of the previous chunk is that starting state of the next. So it doesn't have to go back and look at it again, except to (a) copy it into the `Text` when it's valid and (b) when it's invalid start evaluating the `Word8` at error position + 1 as the start of a code point. Also, the new functions don't construct any new `ByteString`s from the input (with one very small exception).

> 4. Stylistic nit: it's easier to read `case` expressions with disjoint patterns (`Just _ | Nothing`) rather than wildcards (`Just _ | _`), because I don't have to look up previous clauses to find out what constructor is being handled.

4. Removed default (`_ -> ...`) case patterns.

And lastly...

> All of these `append` are going take quadratic time if you try to leniently decode a completely invalid string.

I took into consideration this issue with `streamDecodeUtf8With` in the redesign.

I'm not yet ready to take out of draft status yet, but there's enough to see the approach I'm taking. Also, I rolled back the changes to UTF-16 and 32 decoders for the moment, and just focused on UTF-8.

Lysxia:
Thanks, I'll have a look this weekend.

david-sledge:
Okay, now it's ready for review.

david-sledge (commented):
What's changed to cause Cirrus CI / main to fail with:
```
cabal update
ld-elf.so.1: /lib/libc.so.7: version FBSD_1.7 required by /usr/local/bin/cabal not found
```
Is there anything I can do to resolve this?

Lysxia (commented):
I still think the state is too complicated. You can decompose the decoding so that the core logic only worries about validating UTF-8, and then constructing the `Text` value, i.e., copying buffers, can be implemented on top of that. That core function would have the following signature:

```haskell
parseUtf8Chunk :: ByteString -> Utf8CodePointState -> (Int, Maybe Utf8CodePointState)
-- ^ parseUtf8Chunk chunk s = (n, ms)
-- - When n > 0, n is the length of the longest prefix of chunk that is valid UTF-8 starting from state s
--   (i.e., n points after the end of a full codepoint, to the beginning of an incomplete codepoint).
--  When n = 0, either there is an error, or the whole chunk is a single incomplete codepoint (if s is the starting state), or the chunk is in the middle of an incomplete codepoint (if s is not the starting state).
-- - When ms = Nothing, there is an error: the bytes after index n are not part of a valid UTF-8 code point.
-- - When ms = Just s', all of the remaining bytes after index n are the beginning of some UTF-8 code point, and s' is the corresponding intermediate decoding state, which can be used to parse the next chunk with `parseUtf8Chunk`
```

Various strategies to build `Text` can be implemented on top of it:

- A non-streaming implementation like `decodeUtf8With` just copies valid bytes into a single buffer. 
- A streaming implementation in an arbitrary monad cannot safely use a single mutable buffer, in case the same state is reused with different chunks, so it has to allocate decoded chunks separately.
- But if you can guarantee that state is used linearly, you can still have a streaming decoder with a mutable buffer.

In any case, the core function `parseUtf8Chunk` only needs to tell the decoder is where the valid bytes end.

Also, since an incomplete codepoint is at most 3 bytes, it seems worth carrying it in the state of `parseUtf8Chunk` as well.

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +189,389 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+          )
+      -- no: there's an error
+      Nothing -> huntDownError 0 cpPos cpSt
+
+-- | Decodes a UTF-8 'ByteString' in the context of what has already been
+-- decoded which is represented by the 'Utf8DecodeState' value. Returned is the
+-- new decode state and either ('Right') the number of 'Word8's that make up the
+-- incomplete code point at the end of the input, or ('Left') the start position
+-- of an invalid code point that was encountered. The position is relative to
+-- the start of the input 'ByteString'.
```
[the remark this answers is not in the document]
david-sledge: Yes, it's a little clumsy.

david-sledge:
Prototyped `parseUtf8Chunk` and `parseUtf8ChunkFrom`. The latter would be used for error recovery in the middle of a bytestring. How do they look?

src/Data/Text/Encoding.hs
```diff
@@ -301,6 +296,131 @@ getCodePointStateOrError (Utf8DecodeState mCpSt (bs1@(B.length -> len1) : bss')
...
+ms
+* When ms = Nothing, there is an error: the bytes from index n and beyond are
+  not part of a valid UTF-8 code point.
+* When ms = Just s, all of the remaining bytes from index n and beyond are the
+  beginning of an incomplete UTF-8 code point, and s is the corresponding
+  intermediate decoding state, which can be used to parse the next chunk with
+  `parseUtf8Chunk`
+-}
+parseUtf8ChunkFrom :: ByteString -> Int -> (Int, Maybe Utf8CodePointState)
+parseUtf8ChunkFrom bs@(B.length -> len) pos =
```
Lysxia, line 319: Can start from `0`, assuming the caller used `BS.drop pos` instead?

src/Data/Text/Encoding.hs
```diff
@@ -301,6 +296,131 @@ getCodePointStateOrError (Utf8DecodeState mCpSt (bs1@(B.length -> len1) : bss')
...
+      wc n mask word8 = w n $ (word8 ==) . (mask .&.)
+      guessUtf8Boundary
+        | wi 3 0xf0 = Just $ len - 3  -- third to last char starts a four-byte code point
+        | wi 2 0xe0 = Just $ len - 2  -- pre-last char starts a three-or-four-byte code point
+        | wi 1 0xc2 = Just $ len - 1  -- last char starts a two-(or more-)byte code point
+        | wc 4 0xf8 0xf0 ||           -- last four bytes are a four-byte code point
+          wc 3 0xf0 0xe0 ||           -- last three bytes are a three-byte code point
+          wc 2 0xe0 0xc0 ||           -- last two bytes are a two-byte code point
+          w 1 (< 0x80) = Just len     -- last char is ASCII
+        | otherwise = Nothing         -- no clue
```
Lysxia, line 331: It's only necessary to check for incomplete code points at the end, and not complete ones. After you don't find an incomplete code point, then the whole bytestring is valid or there is an error, and you can guess `len` as the boundary, which will lead to  the necessary validation anyway.

src/Data/Text/Encoding.hs
```diff
@@ -301,6 +296,131 @@ getCodePointStateOrError (Utf8DecodeState mCpSt (bs1@(B.length -> len1) : bss')
...
+#else
+            let bLen = boundary - pos
+                step ndx s
+                  | ndx < pos + bLen =
+                    case updateUtf8State (B.unsafeIndex bs ndx) s of
+                    Just s' -> step (ndx + 1) s'
+                    Nothing -> False
+                  | otherwise = isUtf8StateIsComplete s
+            in
+            step pos utf8StartState
```
Lysxia, line 374: This is can probably be merged with  `huntDownError`

src/Data/Text/Encoding.hs
david-sledge: Yes.

src/Data/Text/Encoding.hs
david-sledge: Done.

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +233,274 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+        )
+      , stackedData'
+      )
+
+-- | Decode a 'ByteString' against a start 'Utf8ValidState' with an empty
+-- 'TextDataStack'.
+--
+-- @decodeUtf8Chunk bs = 'decodeNextUtf8Chunk' bs 'startUtf8ValidState' 'emptyStack'@
+decodeUtf8Chunk :: ByteString -> ((Int, Either (Int, ByteString) Utf8ValidState), TextDataStack)
+decodeUtf8Chunk bs = decodeNextUtf8Chunk bs startUtf8ValidState emptyStack
```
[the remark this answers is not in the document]
david-sledge: A couple of reasons. I didn't want to deal with resizing the buffer, because the exact size of the buffer needed isn't known until the caller request the resulting text (by calling `stackToText`). There's also no guarantee that the caller will request the text, so it saves from having an unnecessary buffer created. Also, I think I ran into the "because type variable ... would escape its scope" error when I tried the buffer.

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +233,274 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
-      | ix >= B.length bs = False
-      -- We do not use decoded code point, so passing a dummy value to save an argument.
-      | otherwise = case utf8DecodeContinue (B.unsafeIndex bs ix) st (CodePoint 0) of
-        Accept{} -> start (ix + 1)
-        Reject{} -> False
-        Incomplete st' _ -> step (ix + 1) st'
+  = validateUtf8 0 0 utf8StartState
+    where
 #endif
+      validateUtf8 ndx0 ndx s
```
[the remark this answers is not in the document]
david-sledge: Will do.

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +233,274 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
 
-isValidBS :: ByteString -> Bool
+-- | This data type represents the state of a 'ByteString' representing
+-- UTF-8-encoded text. It consists of a value representing whether or
+-- not the last byte is a complete code point, and on incompletion what
+-- the 1 to 3 end bytes are that make up the incomplete code point.
+data Utf8ValidState = Utf8ValidState
+  { -- | Get the incomplete UTF-8 code point of the 'ByteString's that
+    -- have been validated thus far.
+    partialUtf8CodePoint :: [ByteString]
```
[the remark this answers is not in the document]
david-sledge: Good point.

src/Data/Text/Encoding.hs
```diff
@@ -135,20 +154,47 @@ import Data.Text.Internal.Encoding.Utf8 (CodePoint(..))
...
-decodeASCII bs = withBS bs $ \fp len -> if len == 0 then empty else runST $ do
-  asciiPrefixLen <- fmap cSizeToInt $ unsafeIOToST $ unsafeWithForeignPtr fp $ \src ->
-    c_is_ascii src (src `plusPtr` len)
-  if asciiPrefixLen == len
-  then let !(SBS.SBS arr) = SBS.toShort bs in
-        return (Text (A.ByteArray arr) 0 len)
-  else error $ "decodeASCII: detected non-ASCII codepoint at " ++ show asciiPrefixLen
+decodeASCII bs =
+  case decodeAsciiPrefix bs of
+    (_, Just errPos) -> error $ "decodeASCII: detected non-ASCII codepoint at " ++ show errPos
```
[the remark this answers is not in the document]
david-sledge: Sounds reasonable.

src/Data/Text/Encoding.hs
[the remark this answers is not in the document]
david-sledge: I started working on this, and I'm wondering if the juice is worth the squeeze.

In the current form, with the head bytestring a reference to it is already being maintained in the `TextDataStack` since `ByteString.drop` and `ByteString.take` simply move the needle without changing the underlying array. With any tail bytestrings, they'll only have a length of one or two words. (Admittedly, this all with the understanding that a bytestring is not just the underlying array, but also a pointer, an integer, and the structure binding the three together, and that each drop and take create more of these structures.)

Additionally, it adds an intermediary translation and the cost of going to and from that translation: `ByteString` -> `Word32` -> `Text`.

How much added value is this for at most three bytestrings whose total length is at most three words?

Bodigrim:
@Lysxia @david-sledge where are we with this? We would have to make a new release to accompany GHC 9.6 soon, it would be great to get this merged.

(Ignore failing drone.io build, it seems they discontinued their free tier service)

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +233,274 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
 #endif
+  ByteString -> Utf8ValidState -> (Int, Either Int Utf8ValidState)
+validateNextUtf8Chunk bs@(B.length -> len) st@(Utf8ValidState lead s)
+  | len > 0 =
+    let g pos s'
+          -- first things first. let's try to get to the start of the next code point
+          | isUtf8StateIsComplete s' =
+            -- found the beginning of the next code point, hand this off to someone else
+            case validateUtf8Chunk $ B.drop pos bs of
+              (len', mS) -> (pos + len', case mS of Left p -> Left (p + pos); _ -> mS)
```
Lysxia, line 356: `Data.Bifunctor.first (+ pos) mS`

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +233,274 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+--   negative if the code point is still incomplete or is proven to be
+--   invalid.
+--   
+-- * The value 'es' indicates whether the 'ByteString'
+--
+--     * (@Left p@) contains an invalid code point and where the next
+--       (potentially valid) code point begins, so that @p - n@ is the
+--       number of invalid bytes, or
+--     * (@Right s'@) is valid, and all of the remaining bytes starting
+--       at inbex 'n' are the beginning of an incomplete UTF-8 code
```
Lysxia, line 341:
```suggestion
--       at index 'n' are the beginning of an incomplete UTF-8 code
```

Lysxia (commented):
I think this is almost good to go. Let's try to finish this within 2-4 weeks.

Bodigrim:
@Lysxia @david-sledge GHC 9.6-alpha1 is out, so it's high time for us to make a release before it's finalised. Ideally I'd like to cut a release by the end of January.

david-sledge:
Demands on my time have left me little time to work on this. However, I'm have another commit in the next 12 to 36 hours.

david-sledge:
Any objects to moving the type class implementations of `Eq Text` and `Ord Text` from `Data.Text` to `Data.Text.Internal`? I want to make instances of `Eq TextDataStack` and `Ord TextDataStack`, but importing `Data.Text` causes a circular dependency.

Lysxia:
Why do you need those instances? I'm not sure we should even expose `TextDataStack`, and instead we should keep it as an implementation detail of `decodeUtf8`. I would prefer the new streaming primitives to only tell the user where the invalid bytes are, so they have full freedom to decide how to do the actual copying. `TextDataStack` requires keeping all chunks in memory so it seems only useful to implement the non-streaming `decodeUtf8`.

david-sledge:
> Why do you need those instances? I'm not sure we should even expose `TextDataStack`, and instead we should keep it as an implementation detail of `decodeUtf8`.

It's not a need. I'd just find it useful for testing and verification. I don't think exposing `TextDataStack` is avoidable, because it's the state value. However, exposure of the data type's details can be minimized, but I don't know what that'd buy us.

> I would prefer the new streaming primitives to only tell the user where the invalid bytes are, so they have full freedom to decide how to do the actual copying. `TextDataStack` requires keeping all chunks in memory so it seems only useful to implement the non-streaming `decodeUtf8`.

So would that be just the validation functions, and not the chunk decoders?

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +235,311 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+-- | Empty stack
+emptyStack :: TextDataStack
+emptyStack = TextDataStack [] 0
+
+-- | Push a text value onto the stack
+pushText :: Text -> TextDataStack -> TextDataStack
+pushText t@(Text _ _ tLen) tds@(TextDataStack stack sLen) =
+  if tLen > 0
+  then TextDataStack (Left t : stack) $ sLen + tLen
+  else tds
```
Lysxia, line 408: This seems to only ever be used to push singletons. Remove it and the `Either Text` from `dataStack`?

Lysxia:
> So would that be just the validation functions, and not the chunk decoders?

Right. I see that the validation functions are already exported, so it's just about removing the other exports.

One way to wrap that into a higher-level API is to view the validation as more of a parser (following the "parse, don't validate" mantra) `Utf8ValidState -> ByteString -> (Text, ByteString, Either Int Utf8ValidState)`, and to avoid the copy incurred by `Text`, you can replace it with some newtype around `Text`.

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +235,311 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+decodeNextUtf8Chunk bs s tds@(TextDataStack stack sLen) =
+  case validateNextUtf8Chunk bs s of
+    (len, res) ->
+      let stackedData'
+            | len >= 0 =
+              let partCP = partialUtf8CodePoint s
+                  partLen = partUtf8CPLen partCP
+                  stackedData@(TextDataStack stack' sLen') =
+                    if partLen > 0
+                    then TextDataStack (Right (Left partCP) : stack) $ sLen + partLen
```
Lysxia, line 480: Isn't `partCP` incomplete here?

src/Data/Text/Encoding.hs
```diff
@@ -187,31 +235,311 @@ decodeLatin1 bs = withBS bs $ \fp len -> runST $ do
...
+validateNextUtf8Chunk ::
+#if defined(ASSERTS)
+  HasCallStack =>
 #endif
+  ByteString -> Utf8ValidState -> (Int, Either Int Utf8ValidState)
+validateNextUtf8Chunk bs@(B.length -> len) st@(Utf8ValidState lead s)
+  | len > 0 =
+    let g pos s'
+          -- first things first. let's try to get to the start of the next code point
+          | isUtf8StateIsComplete s' =
```
Lysxia, line 372: I think the completed codepoint should be part of the output of this function, as a `Maybe Char`.

src/Data/Text/Encoding.hs
david-sledge: This module only uses it to push singletons, but the idea goes beyond how it's implemented here. `pushText` is intended to allow the caller to inject any `Text` onto the stack. Instead of having a single replacement character for invalid code point, it could be a series of characters. Of course it doesn't even have to be at an invalid code point. It can be at any point in between chunk decodings.

Lysxia: I see. I'm not a fan of this data type  It's going to take more iterations to refine the API. Hence I think it's better to reduce the scope of this PR to just the validation functions, which handles the UTF-8-specific logic. This already makes decoding just a matter of slicing the input strings using the indices provided by the validation functions. We can commit to a higher-level interface in a different proposal.

raehik:
I am interested in some parts of this PR which seem standalone from the UTF-8 validation/decoding work, namely the simple safe ASCII bytestring decoding. Is there any way I can help here?

Lysxia:
Sorry for the delay, it took me longer than expected to finalize this. What makes this a hard problem is that there wasn't a clear end goal. So it's worth recapitulating what this is about.

## Incremental decoding

The current API for decoding UTF-8 `ByteString` to `Text` is all or nothing: `decodeUtf8' :: ByteString -> Either Error Text`. For finer-grained error-handling, there is a slightly more general `decodeUtf8With :: (String -> Maybe Word8 -> Maybe Char) -> ByteString -> Text`, but it is still rather limited because the callback gets very little context (a single `Word8`, whereas UTF-8 errors require looking up to 4 byte ahead do diagnose properly) and has very little power (purely return zero or one `Char`). For example, it's not even possible to tell where errors are happening using that API.

This PR is adds new primitives that do less while letting the user do more.

An initial idea was to change `decodeUtf8With` to take a monadic callback, however that turns out not to be feasible efficiently (that would require some way of restricting it to "linear" monads, i.e., not `[]`). Instead we first focus on providing **validation**:

```
validateUtf8Chunk :: ByteString -> (Int, Maybe Utf8State)
validateUtf8More :: Utf8State -> ByteString -> (Int, Maybe Utf8State)
```

You can validate one chunk at a time, threading around the decoder state. Every call gives you an index which delimits a valid UTF-8 chunk of text. All that's left to do for the user is to slice and copy the text, which are quite straightforward operations. Providing just validation means that the user has full freedom in allocating the data structure they want. Maybe they want to allocate one big buffer at the end; maybe they don't even want to allocate anything and just keep the original `ByteString` as the underlying representation (note: this would not be `Text`; `ByteString` is a `ForeignPtr` and `Text` is an `Array`).

With only indices, you still need unsafe operations to slice the `ByteString` and copy it into `Text`. This motivates a more "parse, don't validate" interface for the common case where you do want some `Text` in the end:

```
decodeUtf8Chunk :: ByteString -> (StrictBuilder, ByteString, Maybe Utf8State)
decodeUtf8More :: Utf8State -> ByteString -> (StrictBuilder, ByteString, Maybe Utf8State)
```

Naively, one might have expected `ByteString -> (Text, ByteString, Maybe Utf8State)`. However, that would force you to allocate each chunk separately. `StrictBuilder`, analogous to `Builder` (which is for lazy text), lets the user decide how the final `Text` value is to be built, whether as many chunks or as a single big array.

This also lets us refactor the implementation of UTF-8 decoding nicely. [The previous implementation of `decodeUtf8With`][a] is a monolith of two big nested loops that entangles the core concern of decoding with those of (poor) error handling and streaming. Most of the work in decoding UTF-8 is really in validation, which we've extracted in `validateUtf8Chunk`, so it's now much easier to see what `decodeUtf8With` is doing on top of that.

[a]: https://github.com/haskell/text/blob/8b770aab2a216ac8454939dcdbe36244ae6a0b32/src/Data/Text/Encoding.hs#L244-L317

- [ ] Update Changelog

Boarders:
Very nice write up @Lysxia - makes a ton of sense

src/Data/Text/Encoding.hs
```diff
@@ -68,49 +77,53 @@ module Data.Text.Encoding
...
-import Data.Text.Internal.ByteStringCompat
+import Data.Text.Show ()
 #if defined(ASSERTS)
 import GHC.Stack (HasCallStack)
 #endif
 
 #ifdef SIMDUTF
-import Foreign.C.Types (CInt(..))
 #elif !MIN_VERSION_bytestring(0,11,2)
 import qualified Data.ByteString.Unsafe as B
```
Bodigrim, line 121: Redundant import according to GHA warnings.

src/Data/Text/Internal/Encoding.hs
```diff
@@ -0,0 +1,522 @@
...
+
+#if defined(ASSERTS)
+import Control.Exception (assert)
+#endif
+import Control.Monad.ST (ST, runST)
+import Control.Monad.ST.Unsafe (unsafeIOToST, unsafeSTToIO)
+import Data.Bits ((.&.), shiftL, shiftR)
+import Data.ByteString (ByteString)
+import Data.Functor (void)
+import Data.Semigroup (Semigroup(..))
```
Bodigrim, line 22: This is a redundant import on newer GHCs. There are also redundant imports below.

src/Data/Text/Internal/Encoding.hs
```diff
@@ -0,0 +1,522 @@
...
+{-# LANGUAGE Trustworthy #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- |
+-- Internals of "Data.Text.Encoding".
+--
+-- If you'd like to depend on something from here that's not in "Data.Text.Encoding",
+-- please request to have it exported!
+module Data.Text.Internal.Encoding where
```
Bodigrim, line 12: Please add `@since 2.0.2` to haddock header. Please use an explicit export list. 

Also let's slap a warning about instability / non-adherence to PVP as in other internal modules.

src/Data/Text/Internal/Encoding.hs
```diff
@@ -0,0 +1,522 @@
...
+      | otherwise = (- partUtf8Len part, Just (Utf8State s (partUtf8UnsafeAppend part bs)))
+
+-- | A delayed representation of strict 'Text'.
+
+-- For internal purposes, this is instead used as a delayed 'Array':
+-- it may not actually be valid 'Text' (e.g., 'word8ToStrictBuilder',
+-- 'byteStringToStrictBuilder', 'partUtf8ToStrictBuilder').
+--
+-- @since 2.0.2
+data StrictBuilder = StrictBuilder
```
Bodigrim, line 307: Let's move this to a separate internal, un-exposed module. 

I'm worried that this structure might be too tempting for users to use this strict builder outside of encoding, even while we consider it a very internal thing. Maybe add a strict warning discouraging such usage?

src/Data/Text/Internal/Encoding/Utf8.hs
```diff
@@ -34,19 +34,18 @@ module Data.Text.Internal.Encoding.Utf8
...
     -- * Naive decoding
-    , DecoderResult(..)
-    , DecoderState(..)
-    , CodePoint(..)
-    , utf8DecodeStart
-    , utf8DecodeContinue
+    , Utf8CodePointState
+    , utf8StartState
+    , updateUtf8State
+    , isUtf8StateIsComplete
```
Bodigrim, line 40: While the module is declared as internal, let's nevertheless describe the change in the changelog.

tests/Tests/Properties/Transcoding.hs
```diff
@@ -2,16 +2,22 @@
 
 {-# LANGUAGE OverloadedStrings, ScopedTypeVariables #-}
 {-# OPTIONS_GHC -fno-warn-missing-signatures #-}
+{-# LANGUAGE BangPatterns #-}
 module Tests.Properties.Transcoding
     ( testTranscoding
     ) where
 
+-- import Debug.Trace (trace)
```
Bodigrim, line 10: Clean up?

tests/Tests/Properties/Transcoding.hs
```diff
@@ -2,16 +2,22 @@
...
 import Prelude hiding (head, tail)
 import Data.Bits ((.&.), shiftR)
 import Data.Char (chr, ord)
+import Data.Functor (void)
+import Data.Maybe (isNothing)
+import Data.Semigroup ((<>))
```
Bodigrim, line 16: Redundant import.

tests/Tests/Properties/Transcoding.hs
```diff
@@ -178,6 +351,11 @@ t_decode_utf8_lenient :: Property
 t_decode_utf8_lenient = forAllShrinkShow arbitrary shrink (show . BL.toChunks) $ \bs ->
     decodeLL bs === (TL.fromStrict . decodeL . B.concat . BL.toChunks) bs
 
+-- t_decode_utf8_lenient =
+--   let !res = trace "HEREHEREHERE" $
+--         EL.decodeUtf8With E.lenientDecode (BS.Chunk (B.pack [0xe1]) (BS.Chunk (B.pack [0xa0]) BS.Empty)) === TL.fromStrict (T.pack "\xFFFD\xFFFD") in
+--   trace "THERETHERETHERE" res
```
Bodigrim, line 357: Clean up?

tests/Tests/Properties/Transcoding.hs
```diff
@@ -191,26 +369,126 @@ t_decode_with_error3 =
...
+  let (_pre, suf, ms) = E.decodeUtf8More s chunk
+  in case ms of
+    Nothing -> discard
+    Just s' -> if B.length chunk > B.length suf
+      then p2b s' === suf
+      else p2b s' === p2b s `B.append` suf
+
+-- Decoding chunks separately is equivalent to decoding their concatenation.
+pre_decodeUtf8More_append s chunk1 chunk2 =
+  let (pre1, suf1, ms1) = E.decodeUtf8More s chunk1 in
```
Bodigrim, line 467: Defined but not used: ‘suf1’.

Bodigrim (commented):
Thanks @david-sledge and @Lysxia! Overall looks very good to me, I agree with design decision taken.

src/Data/Text/Internal/Encoding.hs
Lysxia: Should I hide it using CPP? `Semigroup` is needed for base 4.10 which is currently the lower bound for text.

Bodigrim: Yes, unfortunately `CPP` is the way, otherwise there is a flood of warnings from CI.

src/Data/Text/Internal/Encoding.hs
Lysxia: I changed the names of the functions that break this to `unsafeFromByteString` and `unsafeFromWord8`.

src/Data/Text/Internal/Encoding/Utf8.hs
Lysxia: It turns out those were mostly cosmetic changes so I reverted them.

Lysxia (approved):

Lysxia:
I "accepted" my own change to remove the "requested change" flag in the Github UI. Please review :)

I added some optimizations so the benchmarks run at least as fast as before. Most of the time should be spent looping in C++, so there should be no change (confirmed by getting a screenful of "same as baseline" where "baseline" is master), except for the "tiny" benchmark where the work in entering and exiting the function dominates. Somehow the "LazyText" benchmark for the "tiny" input is consistently 25% faster even on my noisy machine. (The strict "Text" benchmark for "tiny" is "same as baseline"; it used to be 3x slower, and that's what my last round of commits fixes).

```All
  Pure
    tiny
      decode
        LazyText: OK (1.20s)
          142  ns ± 7.0 ns, 28% less than baseline
      decode'
        LazyText: OK (0.39s)
          176  ns ± 5.6 ns, 27% less than baseline
```

Note that the benchmarks test valid UTF-8. When there are invalid bytes, the old code performs badly (quadratically) anyway as reported in issue #495. This PR fixes #495.

src/Data/Text/Encoding.hs
```diff
@@ -46,6 +44,16 @@ module Data.Text.Encoding
     , streamDecodeUtf8With
     , Decoding(..)
 
+    -- *** Incremental UTF-8 decoding
+    -- $incremental
+    , decodeUtf8Chunk
+    , decodeUtf8More
+    , Utf8State
+    , startUtf8State
+    , StrictBuilder()
```
Bodigrim, line 53:
```suggestion
    , StrictBuilder
```

changelog.md
```diff
@@ -2,6 +2,31 @@
...
+  * `validateUtf8More`
+
+* Fix quadratic slowdown when decoding invalid UTF-8 bytestrings
+  (https://github.com/haskell/text/issues/495)
+
+* Add internal module `Data.Text.Internal.StrictBuilder`
+
+* Add internal module `Data.Text.Internal.Encoding`
+
+* Add `Data.Text.Internal.Encoding.Utf8.updateDecoderState`
```
Bodigrim, line 28:
```suggestion
* Add `Data.Text.Internal.Encoding.Utf8.updateDecoderState` and export `utf8{Accept,Reject}State` from the same module.
```

Bodigrim (commented):
Just couple of suggestions, otherwise looks very good.

Bodigrim (approved):
Great job!

Lysxia:
Thanks to @david-sledge for getting this started and your great work!

Lysxia:
The test is due to a mismatch between Unicode versions (text = Unicode 14, GHC 8.6 = ???). We probably should disable those tests if the unicode versions don't match.

Bodigrim:
Yeah, `t_toCaseFold_char` fails on `'\66937'`, which is VITHKUQI CAPITAL LETTER FE' (U+10579) added in Unicode 14.0. The test should be guarded with `#if MIN_VERSION_base(4,16,0)`.

@Lysxia could you please rebase?

Lysxia:
Uh, I can if that's what we want to do. But I thought it would be cleaner to squash it. In the Github UI "Rebase and merge" is grayed out but we can still just select "Squash and merge". Is that okay?

Bodigrim:
Ah, I see. Yes, squashing is OK, please go ahead.

414owen:
Hi @david-sledge,

I want to decode some utf8 from a socket, and abort early if I get any invalid data.
The haddock states that the second element of the tuple returned by `decodeUtf8{Chunk,More}` is a 

> undecoded remainder of the given chunk, for diagnosing errors and resuming

It looks like the suffix can also contain input that wasn't parsed because the parser needs more bytes to determine whether it has a valid code point. Is that correct?
If so, {can,how do} I distinguish between a suffix that's due to a unicode error, and one that's due to a lack of input?

Lysxia:
@414owen You can look at the third component of the returned triple `Maybe Utf8State`.

In the doc of `decodeUtf8More`:

> `Just` the new state, or `Nothing` if an invalid byte was encountered (it will be within the first 4 bytes of the undecoded remainder).

PR #455, opened 2022-07-13
Use system-cxx-std-lib virtual package where available

bgamari:
Starting with the 9.4 release, GHC provides `system-cxx-std-lib`, a
"virtual" package which can be used to reliably link against the
system's C++ standard library implementation.

See GHC #20010.


bgamari:
I am still putting this through its paces but the hope is to be able to ship this with 9.4, insulating users (including `text`) from the changes that the Windows toolchain switch brough.

Bodigrim (approved):

PR #466, opened 2022-08-18
Add writeAtMost function to the export list.

Haskell-mouse:
This function gives a possibility to create a Builder with a length unknown initially. 
It is required for double-conversion to create Builder properly when it knows only max length of text but not the final.
Using writeN leads in this case to creating a Builder with max length discarding the actual length of the written buffer. 

The only question I have is if a shrink function should be used inside of
writeAtMost. I am asking because it seems that it only changes "used units" and "length left" but not the actual size of the freed buffer. That could lead to memory over-usage. 
On the other hand, using a shrinking function every time could waste CPU time.

### DIFF
--- a/src/Data/Text/Internal/Builder.hs
+++ b/src/Data/Text/Internal/Builder.hs
@@ -54,6 +54,7 @@ module Data.Text.Internal.Builder
      -- * Internal functions
    , append'
    , ensureFree
+   , writeAtMost
    , writeN
    ) where
 


Bodigrim:
To be honest, I'm a bit reluctant to expose more internals of `Text` `Builder`, because it will make it exponentially more difficult to redesign in future. And given that `text` is a boot package, the added export will not hit the ground until GHC 9.6, so many years will pass before you'll be able to use this function in `double-conversion`.

Are there any workarounds available for your problem?

Haskell-mouse:
> To be honest, I'm a bit reluctant to expose more internals of Text Builder, because it will make it exponentially more difficult to redesign in future.

I understand this, but there probably no workaround which wouldn't lower performance. 
And the performance is the main point of supporting Builder in double-conversion: convertion a large amount of numbers to text via Builder is 10-15 times faster than direct conversion to strict text. 

Ok.... I'll probably try to make a conversion via lazy text.... 

> And given that text is a boot package, the added export will not hit the ground until GHC 9.6, so many years will pass before you'll be able to use this function in double-conversion. 

Oh God, thanks for pointing on this. I'll make one more try to find a way to fast conversion to text. In the worst case I'll redefine a Builder in the double-conversion.

Haskell-mouse:
Another way is to use external builder library like linear-builder

Haskell-mouse:
Ok, I've figured out how to solve a problem using `writeN` with an additional copying of an array. This makes conversion for 5% slower.  
I probably can live with it.

Bodigrim:
Thanks for finding a workaround!

PR #474, opened 2022-10-02
Switch Data.Text.Array to Data.Array.Byte

Bodigrim:
This is a part of a grand plan to unify `ShortByteString`, `Data.Text.Array` and `Data.Primitive.ByteArray` to share the same data type. If it looks good, I'll make a release of `data-array-byte` package.

### DIFF
--- a/.github/workflows/emulated.yml
+++ b/.github/workflows/emulated.yml
@@ -30,8 +30,9 @@ jobs:
         githubToken: ${{ github.token }}
         install: |
           apt-get update -y
-          apt-get install -y ghc libghc-tasty-quickcheck-dev libghc-tasty-hunit-dev
+          apt-get install -y curl ghc libghc-tasty-quickcheck-dev libghc-tasty-hunit-dev
         run: |
+          curl -s https://hackage.haskell.org/package/data-array-byte-0.1/data-array-byte-0.1.tar.gz | tar xz
           ghc --version
-          ghc --make -isrc:tests -o Main cbits/*.c tests/Tests.hs +RTS -s
+          ghc --make -isrc:tests:data-array-byte-0.1 -o Main cbits/*.c tests/Tests.hs +RTS -s
           ./Main +RTS -s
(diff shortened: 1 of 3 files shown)


phadej:
This is technically a breaking change as `Data.Text.Array` is just an ordinary public module, with `Array` being public type.

This will cause problems if someone implemented any instances for `Data.Text.Array.Array`.

Bodigrim:
I hoped no one ever defined instances for `Data.Text.Array`, but https://hackage.haskell.org/package/isomorphism-class-0.1.0.7/docs/IsomorphismClass.html does. Sigh. Well, then a breaking change it is.

phadej:
> I hoped no one ever defined instances for Data.Text.Array

You cannot ever know what (kind of weird stuff) people do in private projects. IMO we cannot assume *no-one* does anything weird, only that is very rare.

Bodigrim:
Cf. https://github.com/haskell/primitive/pull/359 and https://github.com/haskell/bytestring/pull/410.

This is now pending FreeBSD build investigation in #475.

Bodigrim:
Once GHC 9.6 is released, GHC 9.4 will become the oldest bootstrap compiler. So we are free to merge this right after we cut a release for GHC 9.6, which is fairly soon I hope.

Ericson2314:
What's the issue with GHC before 9.4? Hadrian is allowed to depend on things on Hackage.

Bodigrim:
This is ready for review now. Both https://github.com/haskell/primitive/pull/359 and https://github.com/haskell/bytestring/pull/410 have been merged.

Lysxia (approved):

RyanGlScott:
Another breaking change that this PR brings about (aside from the discussion about instances in https://github.com/haskell/text/pull/474#issuecomment-1264711841) is that you will no longer be able to import `Array` like so:

```hs
import Data.Text.Array (Array (..))
```

The `aeson` library actually does this—see [this code](https://github.com/haskell/aeson/blob/274b562695863dc4a1b2e9a8c9d174c2b78cdd12/src/Data/Aeson/Internal/Text.hs#L11). As a result, `aeson` no longer builds with the latest `text` `master` branch, which is slated to be released with GHC 9.8. (See [this failing GHC 9.8 `head.hackage` CI job](https://gitlab.haskell.org/ghc/head.hackage/-/jobs/1609824#L8017).) I only bring this up because the `text` version is still `2.0.2` (the same version released with GHC 9.6), and it took me a while to figure out what had changed.

PR #481, opened 2022-12-10
ci: Fix run-haddock flag

Lysxia:
I typoed the variable name so haddock was never run, and I didn't find a way to have a default value when the variable is unset, so I had to set it for each job.


Bodigrim:
@Lysxia it's better to add `haddock: >=8.6` to `cabal.haskell-ci` and regenerate `haskell-ci.yml`  with `haskell-ci 'github' '--config=cabal.haskell-ci' 'text.cabal'`.

Bodigrim (approved):

PR #482, opened 2022-12-11
Declare pure FFI functions as not living in IO

sergv:
# Synopsis

In some places `text` uses `unsafeDupablePerformIO` which forces allocation of thunk for its result. Most of those cases are for foreign functions which are morally equivalent to C’s `sin` (sine calculation) function in the sense that they’re actually pure.

This PR removes the thunk mentioned by not introducing `IO` in those cases.

## Thunk discussion

Due to some recent-ish fix `unsafeDupablePerformIO` (https://gitlab.haskell.org/ghc/ghc/-/issues/19181) the function now uses `GHC.Magic.lazy` which forces thunk allocation and no optimization will ever make it cease to do so.

For instance, with current implementation of `length` for text the resulting length is always wrapped into a thunk even if demanded strictly.

For example, next is the core of the worker for the function `foo` that takes text’s length.

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends:
  , base
  , text > 2.0
-}

{-# LANGUAGE BangPatterns #-}

{-# OPTIONS_GHC -O2 #-}
{-# OPTIONS_GHC -ddump-simpl -dsuppress-uniques -dsuppress-idinfo -dsuppress-module-prefixes -dsuppress-type-applications -dsuppress-coercions -dppr-cols200 -dsuppress-type-signatures -ddump-to-file #-}

module Main (foo, main) where

import Data.Text (Text)
import qualified Data.Text.IO as T

foo :: Text -> Int
foo !x = len + len
  where
    !len = T.length x

main :: IO ()
main =
  print =<< foo <$> T.getContents

```

By running `cabal run ./Main.hs <Main.hs` with GHC 9.4.3:

```haskell
-- RHS size: {terms: 31, types: 20, coercions: 3, joins: 0/0}
$wfoo
  = \ ww ww1 ww2 ->
      case ww2 of wild1 {
        __DEFAULT ->
          runRW#
            (\ s ->
               case {__ffi_static_ccall_unsafe text-2.0.1:_hs_text_measure_off :: ByteArray#
                                                              -> Word64#
                                                              -> Word64#
                                                              -> Word64#
                                                              -> State# RealWorld
                                                              -> (# State# RealWorld, Int64# #)}
                      ww (int64ToWord64# (intToInt64# ww1)) (int64ToWord64# (intToInt64# wild1)) 9223372036854775807##64 s
               of
               { (# ds, ds1 #) ->
               case (lazy ((I64# ds1) `cast` <Co:2> :: Int64 ~R# CSsize)) `cast` <Co:1> :: CSsize ~R# Int64 of { I64# x1 ->
               *# -2# (int64ToInt# x1)
               }
               });
        0# -> 0#
      }
```

Result of `_hs_text_measure_off` is lazily allocate before bein consumed.

## Ways to remove thunk

In order to remove the extra thunk the `Data.Bytestring.Internal.accursedUnutterablePerformIO` function could be used, or its cousing in the `text` package `Data.Text.Internal.Unsafe.inlinePerformIO`. Their definitions are equivalent so they should be equally unsafe, accursed and unutterable.

The accursedness and unutterability of those functions mosty comes from their interaction with GHC’s optimizer. When arbitrary IO action is supplied to them some invariants may be violated. But when a FFI wrapper like `_hs_text_measure_off` is supplied then it’s not so bad since GHC never sees inside the wrapper.

However with FFI wrappers it’s not mandatory to make them operate in IO. C’s wrapper for trigonometric sine caculation, `sin`, is declared pure because it’s pure and doesn’t have side effects. IO is important if the wrapped function is going to become part of bigger IO action where ordering of effects doesn’t matter. I argue that functions like `_hs_text_measure_off` in the text package are actually like C’s `sin` and may be declared pure from the start, which this PR does to some extent.

By going this route there’s not need for any variety of unsafe perform IO since there’s no IO.

As a side effect of this PR the `Data.Text.Internal.Unsafe.inlinePerformIO` can probably be deprecated at some point in the future when other two uses of it in the `Data.Text.Internal.IO` module are dealt with.

### DIFF
--- a/src/Data/Text/Internal/Lazy/Search.hs
+++ b/src/Data/Text/Internal/Lazy/Search.hs
@@ -66,7 +66,7 @@ indices needle
          c = index xxs (i + nlast)
          delta | nextInPattern = nlen + 1
                | c == z        = skip + 1
-               | l >= i + nlen = case unsafeDupablePerformIO $
+               | l >= i + nlen = case
                   memchr xarr# (intToCSize (xoff + i + nlen)) (intToCSize (l - i - nlen)) z of
                     -1 -> max 1 (l - i - nlen)
                     s  -> cSsizeToInt s + 1
@@ -142,4 +142,4 @@ cSsizeToInt :: CSsize -> Int
 cSsizeToInt = fromIntegral
 
 foreign import ccall unsafe "_hs_text_memchr" memchr
-    :: ByteArray# -> CSize -> CSize -> Word8 -> IO CSsize
+    :: ByteArray# -> CSize -> CSize -> Word8 -> CSsize
--- a/src/Data/Text/Internal/Lazy/Search.hs
+++ b/src/Data/Text/Internal/Lazy/Search.hs
@@ -32,7 +32,6 @@ import qualified Data.Text.Internal as T
 import qualified Data.Text as T (concat, isPrefixOf)
 import Data.Text.Internal.Fusion.Types (PairS(..))
 import Data.Text.Internal.Lazy (Text(..), foldrChunks)
-import Data.Text.Unsafe (unsafeDupablePerformIO)
 import Data.Bits ((.|.), (.&.))
 import Foreign.C.Types
 import GHC.Exts (ByteArray#)
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1371,13 +1371,13 @@ take n t@(Text arr off len)
 -- @since 2.0
 measureOff :: Int -> Text -> Int
 measureOff !n (Text (A.ByteArray arr) off len) = if len == 0 then 0 else
-  cSsizeToInt $ unsafeDupablePerformIO $
+  cSsizeToInt $
     c_measure_off arr (intToCSize off) (intToCSize len) (intToCSize n)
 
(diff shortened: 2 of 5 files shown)


Bodigrim:
Makes sense to me, I don't know why these FFI functions were operating in IO in the first place.

Bodigrim:
@sergv could you please fix build warnings for redundant imports?

src/Data/Text/Internal/Lazy/Search.hs
```diff
@@ -66,7 +66,7 @@ indices needle
          c = index xxs (i + nlast)
          delta | nextInPattern = nlen + 1
                | c == z        = skip + 1
-               | l >= i + nlen = case unsafeDupablePerformIO $
+               | l >= i + nlen = case
```
Bodigrim, line 69: The import of `Data.Text.Unsafe` is redundant now.

sergv:
My bad, somewhat hoped CI will fail with -Werror on everything. To check older GHCs I'd still rely on CI anyway.

Bodigrim (approved):

Lysxia (approved):

Issue #479, opened 2022-12-08
UTF8 decode on unpinned bytes


andrewthad:
I have, in the `byteslice` library, a type that looks like this:

```
data Bytes = Bytes
  { array :: {-# UNPACK #-} !ByteArray
  , offset :: {-# UNPACK #-} !Int
  , length :: {-# UNPACK #-} !Int
  }
```

This is the same thing as `ByteString` except that it doesn't require pinned memory and it cannot use memory that was allocated in C code. I'm trying to write this function (not in `text`, in my library):

    decodeUtf8Bytes :: Text -> Maybe Bytes

The `text` library comes with a fast utf8 validation routine implemented in C++. However, it does not expose this in a way that lets me use it. To expose this, it would be sufficient to add this to `text`:

```
/* Add this to cbits/validate_utf8.cpp */
extern "C"
int _hs_text_is_valid_utf8_offset(const char* str, size_t off, size_t len){
  return simdutf::validate_utf8(str + off, len);
}
```

And a wrapper:

```
foreign import ccall unsafe "_hs_text_is_valid_utf8_offset" c_is_valid_utf8_offset
    :: ByteArray# -> CSize -> CSize -> IO CInt
```

With this wrapper, it becomes possible to perform UTF-8 validation of unpinned `ByteArray#` at arbitrary starting points.

If something like this were added to `text`, it could be exposed in an internal, unstable module. Let me know if this sounds like a welcome addition (and if it is, with some direction on where this should be exposed), and I can prepare a patch.

Boarders:
I think this would be a welcome addition (especially as it will only be promised internally).

Bodigrim:
Looks reasonable to me.

phadej:
You probably want both `ccall safe` and `ccall unsafe` variants of `_hs_text_is_valid_utf8_offset`.

For big enough `ByteArray#` it will be pinned, so it might be good idea to check whether length is big enough, then check that array is actually pinned, and go the `safe` route.

andrewthad:
I've opened a PR with this at https://github.com/haskell/text/pull/483.

One thing I realized as I was doing this is that I need to provide a fallback when the SIMDUTF flag is off. I need to add a variant of the `isValidBS` fallback that works on `ByteArray#` instead of `ByteString`. I don't think this is terribly difficult, but I've not done it yet.

andrewthad:
I've added the important missing stuff.

Bodigrim:
Closed by #483.

PR #483, opened 2022-12-27
Expose SIMD UTF-8 validation functions from internal module

andrewthad:
This makes it possible for users to validate ByteArray# when compiling with the SIMDUTF8 flag on. It does not, however, provide any fallbacks for when that flag is off. That is left as future work.

Related to https://github.com/haskell/text/issues/479

### DIFF
--- a/src/Data/Text/Internal/PrimCompat.hs
+++ b/src/Data/Text/Internal/PrimCompat.hs
@@ -13,13 +13,12 @@ module Data.Text.Internal.PrimCompat
   ) where
 
 #if MIN_VERSION_base(4,16,0)
-
-import GHC.Base
-
+import GHC.Exts (wordToWord8#,word8ToWord#,wordToWord16#,word16ToWord#,wordToWord32#,word32ToWord#)
 #else
-
 import GHC.Prim (Word#)
+#endif
 
+#if !(MIN_VERSION_base(4,16,0))
 wordToWord8#,  word8ToWord#  :: Word# -> Word#
 wordToWord16#, word16ToWord# :: Word# -> Word#
 wordToWord32#, word32ToWord# :: Word# -> Word#
@@ -33,5 +32,4 @@ wordToWord32# w = w
 {-# INLINE word16ToWord# #-}
 {-# INLINE wordToWord32# #-}
 {-# INLINE word32ToWord# #-}
-
 #endif
--- a/src/Data/Text/Internal/Validate.hs
+++ b/src/Data/Text/Internal/Validate.hs
@@ -28,7 +28,6 @@ module Data.Text.Internal.Validate
   ) where
 
 import Data.Array.Byte (ByteArray(ByteArray))
-import GHC.Exts (ByteArray#)
 import Data.ByteString (ByteString)
 import GHC.Exts (isTrue#,isByteArrayPinned#)
 
@@ -38,6 +37,7 @@ import Data.Text.Internal.ByteStringCompat (withBS)
 import Data.Text.Internal.Unsafe (unsafeWithForeignPtr)
 import Data.Text.Internal.Validate.Simd (c_is_valid_utf8_bytearray_safe,c_is_valid_utf8_bytearray_unsafe,c_is_valid_utf8_ptr_unsafe)
 #else
+import GHC.Exts (ByteArray#)
 import Data.Text.Internal.Encoding.Utf8 (CodePoint(..),DecoderResult(..),utf8DecodeStart,utf8DecodeContinue)
 import GHC.Exts (Int(I#),indexWord8Array#)
 import GHC.Word (Word8(W8#))
--- a/src/Data/Text/Internal/Validate.hs
+++ b/src/Data/Text/Internal/Validate.hs
@@ -1,0 +1,149 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnliftedFFITypes #-}
+
+-- | Test whether or not a sequence of bytes is a valid UTF-8 byte sequence.
+-- In the GHC Haskell ecosystem, there are several representations of byte
+-- sequences. The only one that the stable @text@ API concerns itself with is
+-- 'ByteString'. Part of bytestring-to-text decoding is 'isValidUtf8ByteString',
+-- a high-performance UTF-8 validation routine written in C++ with fallbacks
+-- for various platforms. The C++ code backing this routine is nontrivial,
+-- so in the interest of reuse, this module additionally exports functions
+-- for working with the GC-managed @ByteArray@ type. These @ByteArray@
+-- functions are not used anywhere else in @text@. They are for the benefit
+-- of library and application authors who do not use 'ByteString' but still
+-- need to interoperate with @text@.
+module Data.Text.Internal.Validate
+  (
+  -- * ByteString
+    isValidUtf8ByteString
+  -- * ByteArray
+  --
+  -- | Is the slice of a byte array a valid UTF-8 byte sequence? These
+  -- functions all accept an offset and a length.
+  , isValidUtf8ByteArray
+  , isValidUtf8ByteArrayUnpinned
+  , isValidUtf8ByteArrayPinned
+  ) where
+
+import Data.Array.Byte (ByteArray(ByteArray))
+import GHC.Exts (ByteArray#)
+import Data.ByteString (ByteString)
+import GHC.Exts (isTrue#,isByteArrayPinned#)
+
+#ifdef SIMDUTF
+import Data.Text.Unsafe (unsafeDupablePerformIO)
+import Data.Text.Internal.ByteStringCompat (withBS)
+import Data.Text.Internal.Unsafe (unsafeWithForeignPtr)
+import Data.Text.Internal.Validate.Simd (c_is_valid_utf8_bytearray_safe,c_is_valid_utf8_bytearray_unsafe,c_is_valid_utf8_ptr_unsafe)
+#else
+import Data.Text.Internal.Encoding.Utf8 (CodePoint(..),DecoderResult(..),utf8DecodeStart,utf8DecodeContinue)
+import GHC.Exts (Int(I#),indexWord8Array#)
+import GHC.Word (Word8(W8#))
+import qualified Data.ByteString as B
+#if !MIN_VERSION_bytestring(0,11,2)
+import qualified Data.ByteString.Unsafe as B
+#endif
+#endif
+
+-- | Is the ByteString a valid UTF-8 byte sequence?
+isValidUtf8ByteString :: ByteString -> Bool
+#ifdef SIMDUTF
+isValidUtf8ByteString bs = withBS bs $ \fp len -> unsafeDupablePerformIO $
+  unsafeWithForeignPtr fp $ \ptr -> (/= 0) <$> c_is_valid_utf8_ptr_unsafe ptr (fromIntegral len)
+#else
+#if MIN_VERSION_bytestring(0,11,2)
+isValidUtf8ByteString = B.isValidUtf8
+#else
+isValidUtf8ByteString bs = start 0
+  where
+    start ix
+      | ix >= B.length bs = True
+      | otherwise = case utf8DecodeStart (B.unsafeIndex bs ix) of
+        Accept{} -> start (ix + 1)
+        Reject{} -> False
+        Incomplete st _ -> step (ix + 1) st
+    step ix st
+      | ix >= B.length bs = False
+      -- We do not use decoded code point, so passing a dummy value to save an argument.
+      | otherwise = case utf8DecodeContinue (B.unsafeIndex bs ix) st (CodePoint 0) of
+        Accept{} -> start (ix + 1)
+        Reject{} -> False
+        Incomplete st' _ -> step (ix + 1) st'
+#endif
+#endif
+
+-- | For pinned byte arrays larger than 128KiB, this switches to the safe FFI
+-- so that it does not prevent GC. This threshold (128KiB) was chosen
+-- somewhat arbitrarily and may change in the future.
+isValidUtf8ByteArray ::
+     ByteArray -- ^ Bytes
+  -> Int -- ^ Offset
+  -> Int -- ^ Length
+  -> Bool
+isValidUtf8ByteArray b@(ByteArray b#) !off !len
+  | len >= 131072 -- 128KiB
+  , isTrue# (isByteArrayPinned# b#)
+  = isValidUtf8ByteArrayPinned b off len
+  | otherwise = isValidUtf8ByteArrayUnpinned b off len
+
+-- | This uses the @unsafe@ FFI. GC waits for all @unsafe@ FFI calls
+-- to complete before starting. Consequently, an @unsafe@ FFI call does not
+-- run concurrently with GC and is not interrupted by GC. Since relocation
+-- cannot happen concurrently with an @unsafe@ FFI call, it is safe
+-- to call this function with an unpinned byte array argument.
+-- It is also safe to call this with a pinned @ByteArray@ argument.
+isValidUtf8ByteArrayUnpinned ::
+     ByteArray -- ^ Bytes
+  -> Int -- ^ Offset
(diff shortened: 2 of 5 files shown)


andrewthad:
I've added portable functions that work without SIMDUTF being turned on. This is ready for review.

andrewthad:
Fixed a problem with GHC 8.0. I still need to add some tests to the test suite. The `ByteArray#` functions have no coverage.

andrewthad:
Just squashed this down to remove a bunch of meaningless commits.

src/Data/Text/Internal/PrimCompat.hs
```diff
@@ -10,16 +10,23 @@ module Data.Text.Internal.PrimCompat
...
-
-import GHC.Base
-
+import GHC.Exts (wordToWord8#,word8ToWord#,wordToWord16#,word16ToWord#,wordToWord32#,word32ToWord#)
 #else
-
 import GHC.Prim (Word#)
+#endif
 
+#if MIN_VERSION_base(4,10,0)
```
Bodigrim, line 23: I think it's time for `base-4.9` (and corresponding GHC 8.0) to go: I'd rather drop its support than accumulate more boilerplate. @Lysxia what do you think?

Lysxia: I also think we can drop base-4.9.

andrewthad: I’m fine with killing off GHC 8.0 support. Should I open a separate PR doing that? I think it would be best for that to be a separate commit and not tied up in this addition.

Bodigrim: Yes, please make a separate PR to decomission GHC 8.0.

andrewthad: Done.

andrewthad:
This is ready for another round of review now that support for GHC 8.0 is gone. The only changes since last time are documentation. I've provided the motivation for the module in the module's haddock, and I've documented the behaviors of the functions.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,155 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnliftedFFITypes #-}
+#if defined(SIMDUTF) || MIN_VERSION_bytestring(0,11,2)
+{-# LANGUAGE Trustworthy #-}
```
Bodigrim, line 6: `Data.Text.Encoding` used to promise `Trustworthy` unconditionally. Any particular reason not to follow its example?

I would rather remove `{-# LANGUAGE Trustworthy #-}` (not a fan of "Safe" Haskell) from both modules than promise it only conditionally, because it is likely to explode somewhere later in the ecosystem.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,155 @@
...
+import Data.ByteString (ByteString)
+import GHC.Exts (isTrue#,isByteArrayPinned#)
+
+#ifdef SIMDUTF
+import Data.Text.Unsafe (unsafeDupablePerformIO)
+import Data.Text.Internal.ByteStringCompat (withBS)
+import Data.Text.Internal.Unsafe (unsafeWithForeignPtr)
+import Data.Text.Internal.Validate.Simd (c_is_valid_utf8_bytearray_safe)
+import Data.Text.Internal.Validate.Simd (c_is_valid_utf8_bytearray_unsafe)
+import Data.Text.Internal.Validate.Simd (c_is_valid_utf8_ptr_unsafe)
```
Bodigrim, line 48: Please collapse these three imports.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,155 @@
...
+#else
+import Data.Text.Internal.Encoding.Utf8 (CodePoint(..),DecoderResult(..))
+import Data.Text.Internal.Encoding.Utf8 (utf8DecodeStart,utf8DecodeContinue)
```
Bodigrim, line 51: And these two.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,155 @@
...
+import GHC.Exts (Int(I#),indexWord8Array#)
+import GHC.Word (Word8(W8#))
+#if MIN_VERSION_bytestring(0,11,2)
+import qualified Data.ByteString as B
+#else
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Unsafe as B
```
Bodigrim, line 58: Let's `import qualified Data.ByteString as B` unconditionally.

text.cabal
```diff
@@ -170,6 +170,8 @@ library
     Data.Text.Internal.Search
     Data.Text.Internal.Unsafe
     Data.Text.Internal.Unsafe.Char
+    Data.Text.Internal.Validate
+    Data.Text.Internal.Validate.Simd
```
Bodigrim, line 174: Please make this module conditional: `if flag(simdutf) exposed-modules: Data.Text.Internal.Validate.Simd`

src/Data/Text/Internal/Validate.hs
andrewthad: I'll just remove it. I'm not even sure what guarantees Safe Haskell is supposed to provide, but it's trivial to trigger a segfault by misusing the functions exposed by this module.

src/Data/Text/Internal/Validate.hs
andrewthad: Will do.

src/Data/Text/Internal/Validate.hs
andrewthad: Will do.

text.cabal
andrewthad: Will do. Also, I'll remove the conditionals from the `Simd` module itself since it's compilation will be guarded in this way.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,147 @@
...
+  -> Int -- ^ Offset
+  -> Int -- ^ Length
+  -> Bool
+isValidUtf8ByteArray# b !off !len
+  | len >= 131072 -- 128KiB
+  , isTrue# (isByteArrayPinned# b)
+  = isValidUtf8ByteArraySafe# b off len
+  | otherwise = isValidUtf8ByteArrayUnsafe# b off len
+
+-- | This always uses the unsafe FFI. It is always safe to call this
```
Bodigrim, line 95: I understand what you mean, but I imagine plenty of people puzzled why it is always safe to call unsafe FFI. Could you please elaborate a little bit more, "the unsafe FFI, which is guaranteed not to be interrupted by GC" or similar?

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,147 @@
...
+#ifdef SIMDUTF
+isValidUtf8ByteArrayUnsafe# bs !off !len =
+  unsafeDupablePerformIO $ (/= 0) <$> c_is_valid_utf8_bytearray_unsafe bs (fromIntegral off) (fromIntegral len)
+#else
+isValidUtf8ByteArrayUnsafe# = isValidUtf8ByteArrayHaskell#
+#endif
+
+-- | This always uses the safe FFI. The byte array /must/ be pinned,
+-- and the calling context is responsible for enforcing this. If the
+-- byte array is not pinned, this function's behavior is undefined.
```
Bodigrim, line 111: Same as above, let's elaborate for less experienced users about the interaction between safe FFI, pinnedness and GC.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,147 @@
...
+-- so in the interest of reuse, this module additionally exports functions
+-- for working with the GC-managed @ByteArray#@ type. These @ByteArray#@
+-- functions are not used anywhere else in @text@. They are for the benefit
+-- of library and application authors who do not use 'ByteString' but still
+-- need to interoperate with @text@.
+--
+-- Note: The release of GHC 9.4 has seen a coordinated move of the lifted
+-- @ByteArray@ type from @primitive@ to @base@. In many years, when support
+-- for GHC older than 9.4 is dropped, it will be possible to add functions
+-- for the lifted ByteArray type as well.
```
Bodigrim, line 21: Please rebase your branch atop of https://github.com/haskell/text/pull/474 and switch types to a lifted `type Array = ByteArray` from `Data.Text.Array`. I'm a bit reluctant to introduce API for unlifted arrays just to be painfully deprecated (or extended) soon.

src/Data/Text/Internal/Validate.hs
andrewthad: I've elaborated more in latest commit.

src/Data/Text/Internal/Validate.hs
Bodigrim: I have no strong preferences here.

andrewthad:
Failure on GHC 8.6.5 looks spurious.

andrewthad:
Just squashed. This will kick off another round of CI.

andrewthad:
CI is still failing, but I’m pretty sure that it’s spurious. This is ready for another round of review

andrewthad:
Wait, something really is wrong on the ppcle64 architecture. For some reason, the dependency on the byte array compatibility library doesn’t get picked up, and so it ends up missing the module. I don’t see anything wrong in the Cabal file though.

Bodigrim:
@andrewthad necessary changes are in https://github.com/haskell/text/pull/474/commits/d0b70f782b22ca3d39c35b3cbc042757340cf8b1. I think you squashed a bit overzealously.

andrewthad:
I’ve incorporated the diff manually. Thanks. Looks like that fixed it.

andrewthad:
Only remaining CI failure seems unrelated to the change.

Bodigrim:
It's good to go, but on hold waiting for #474 to land (which in its turn waits until we cut a release to accompany GHC 9.6).

Bodigrim:
@Lysxia given that #448 has landed, could you take a look at what remains of this PR? I assume the new design is flexible enough already, but we might still want to expose low-level calls to simdutf?..

I’m AFK and cannot take a close look myself.

andrewthad:
Is there anything I can do to move this forward? Should I resolve the merge conflict, or are there concerns about this feature no longer fitting into the design of `text`?

Lysxia:
We can still accept this. The bytearray validation functions are new. And while we now have a fancier bytestring validation function, adding the bool version would look consistent with the bytearray ones.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,149 @@
...
+#else
+isValidUtf8ByteArrayUnsafe (ByteArray bs) = isValidUtf8ByteArrayHaskell# bs
+#endif
+
+-- | This uses the @safe@ FFI. GC may run concurrently with @safe@
+-- FFI calls. Consequently, unpinned objects may be relocated while a
+-- @safe@ FFI call is executing. The byte array argument /must/ be pinned,
+-- and the calling context is responsible for enforcing this. If the
+-- byte array is not pinned, this function's behavior is undefined.
+isValidUtf8ByteArraySafe ::
```
Lysxia, line 114: Instead of `Unsafe/Safe` why not `Unpinned/Pinned`?

The `Safe/Unsafe` naming is confusing since "safe FFI" calls are unsafe and "unsafe FFI" calls are safe. The FFI functions themselves don't need to be renamed.

We could also consider `Unsafe/Pinned` but this still gives the wrong impression that the pinned one is safer than `Unsafe`.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,149 @@
...
+  , isTrue# (isByteArrayPinned# b#)
+  = isValidUtf8ByteArraySafe b off len
+  | otherwise = isValidUtf8ByteArrayUnsafe b off len
+
+-- | This uses the @unsafe@ FFI. GC waits for all @unsafe@ FFI calls
+-- to complete before starting. Consequently, an @unsafe@ FFI call does not
+-- run concurrently with GC and is not interrupted by GC. Since relocation
+-- cannot happen concurrently with an @unsafe@ FFI call, it is safe
+-- to call this function with an unpinned byte array argument.
+-- (It is also safe to call this with a pinned @ByteArray@ argument.)
```
Lysxia, line 96:
```suggestion
-- It is also safe to call this with a pinned @ByteArray@ argument.
```

That's really a proper part of the spec; it's even already relied upon by the other function.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,149 @@
...
+      | ix >= B.length bs = False
+      -- We do not use decoded code point, so passing a dummy value to save an argument.
+      | otherwise = case utf8DecodeContinue (B.unsafeIndex bs ix) st (CodePoint 0) of
+        Accept{} -> start (ix + 1)
+        Reject{} -> False
+        Incomplete st' _ -> step (ix + 1) st'
+#endif
+#endif
+
+-- | For pinned byte arrays larger than 128KiB, this switches to the safe FFI
```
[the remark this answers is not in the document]
andrewthad: I'm going to answer this on the main thread, and we can discuss it there. I started writing this, and I realized there's enough information that I don't want github's comment folding to make it impossible for future readers to find it.

Lysxia:
That's a very clear analysis! That makes sense to me. Confirming this further, the GHC User Manual also says:

> This means that if you need to make a foreign call to a function that takes a long time or potentially blocks, then you should mark it safe and use -threaded. Some library functions make such calls internally; their documentation should indicate when this is the case.
>
> On the other hand, a foreign call to a function that is guaranteed to take a short time, and does not call back into Haskell can be marked unsafe.

Bodigrim:
@andrewthad could you please rebase?

andrewthad:
I've just now squashed everything down and rebased.

src/Data/Text/Internal/Validate.hs
```diff
@@ -0,0 +1,149 @@
...
+  --
+  -- | Is the slice of a byte array a valid UTF-8 byte sequence? These
+  -- functions all accept an offset and a length.
+  , isValidUtf8ByteArray
+  , isValidUtf8ByteArrayUnpinned
+  , isValidUtf8ByteArrayPinned
+  ) where
+
+import Data.Array.Byte (ByteArray(ByteArray))
+import GHC.Exts (ByteArray#)
```
Bodigrim, line 31: Redundant import at least for recent GHCs.

text.cabal
```diff
@@ -136,6 +137,10 @@ library
   if flag(simdutf) && os(netbsd) && impl(ghc < 9.4)
     build-depends: base < 0
 
+  -- Before GHC 9.4, Data.Array.Byte is not in base
+  if impl(ghc < 9.4)
+    build-depends: data-array-byte >= 0.1.0.1
```
Bodigrim, line 142: This is redundant, `build-depends: data-array-byte` is already in `text.cabal`.

text.cabal
```diff
@@ -85,6 +85,7 @@ library
   hs-source-dirs: src
 
   if flag(simdutf)
+    exposed-modules: Data.Text.Internal.Validate.Simd
```
Bodigrim, line 88: Please mention this additional module in the flag description above.

src/Data/Text/Internal/Validate/Simd.hs
```diff
@@ -0,0 +1,43 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
```
Bodigrim, line 2: Seems there is no CPP used?

src/Data/Text/Internal/Validate.hs
andrewthad: I've now guarded this behind CPP checking that simdutf is not defined.

Bodigrim (approved):

andrewthad:
I'm not able to make sense of the NetBSD test failure output.

andrewthad:
Is there anything else that I can improve in this PR, or is this ready to be merged?

Lysxia (approved):

PR #491, opened 2023-01-19
pack: shrink the backing mutable byte array after 'Chars' have been added

sidkshatriya:
The `pack` function converts a `String` to `Text`. It allocates 4 bytes per `Char` into a mutable byte array (as a UTF-8 character will take 4 bytes in the worse case). After all the `Char`s have been written to the byte array, we should shrink the mutable byte array to the actual size required.

This shrinking operation does not seem to be present. This PR adds it. This should lead to substantial memory savings for long `String`s.

Sorry if I've not understood something. Let me know if the `shrinkM` is not required.

### DIFF
--- a/src/Data/Text/Internal.hs
+++ b/src/Data/Text/Internal.hs
@@ -251,6 +251,10 @@ pack xs = runST $ do
         d <- unsafeWrite marr off (safe c)
         go (off + d) cs
   len <- go 0 xs
+  -- On average, the number of bytes taken by a typical 'Char' is less than 4 in UTF-8
+  -- (only 1 to 2 bytes for latin-script languages for instance), so we shrink the array
+  -- now. Could mean substantial memory savings for a long 'String'.
+  A.shrinkM marr len
   arr <- A.unsafeFreeze marr
   return (Text arr 0 len)
 {-# NOINLINE [0] pack #-}


Lysxia (approved):
Looks reasonable to me!

Bodigrim (approved):
Thanks!

sidkshatriya:
@Bodigrim Friendly ping! Will greatly appreciate if you could have a look...

sidkshatriya:
@Bodigrim Thanks for the merge. 

I don't have a feeling for how frequently the `pack` function is called on `String`s in an typical program using `text` and therefore how much memory we might save in practice. Are there any benchmarking suites available to compare before/after this patch?

Might it make sense to have tag a 2.0.1.1 or 2.0.2 incorporating this change?

Bodigrim:
Unfortunately, here are several loose ends to tighten up before a release. If you want to speed it up, how do you feel about tackling this TODO?
https://github.com/haskell/text/blob/8b770aab2a216ac8454939dcdbe36244ae6a0b32/src/Data/Text/Internal.hs#L261

Even after your fix in this PR `pack` function allocates the input list in full (because it is forced by `length`) + up to 4x space for a bytearray. All this is freed afterwards, but even temporary memory spike could be harmful. This function should be rewritten to start with a very small bytearray, which is resized while consuming the list of characters. A similar problem is dealt with for `filter`:
https://github.com/haskell/text/blob/8b770aab2a216ac8454939dcdbe36244ae6a0b32/src/Data/Text.hs#L1758-L1760

PR #497, opened 2023-01-26
add C-based `isAscii :: Text -> Bool`

raehik:
Now that `Text`s are represented internally as a UTF-8 bytestring, we can provide a fast `isAscii :: Text -> Bool` that inspects the bytestring directly, rather than `Char` by `Char`. Such a function can come in handy in serialization libraries, and can't easily be written by an end user. Plus, the C snippet already exists, so there's not much to do.

### DIFF
--- a/cbits/is_ascii.c
+++ b/cbits/is_ascii.c
@@ -48,5 +48,5 @@ const size_t _hs_text_is_ascii(const uint8_t *src0, const uint8_t *srcend){
 
 // TODO wrapper for using with ByteArray#
 const size_t _hs_text_is_ascii_2(const uint8_t *arr, size_t off, size_t len){
-    return _hs_text_is_ascii(arr + off, arr + len);
+    return _hs_text_is_ascii(arr + off, arr + off + len);
 }
--- a/cbits/is_ascii.c
+++ b/cbits/is_ascii.c
@@ -45,3 +45,8 @@ const size_t _hs_text_is_ascii(const uint8_t *src0, const uint8_t *srcend){
 
   return src - src0;
 }
+
+// TODO wrapper for using with ByteArray#
+const size_t _hs_text_is_ascii_2(const uint8_t *arr, size_t off, size_t len){
+    return _hs_text_is_ascii(arr + off, arr + len);
+}
--- a/cbits/is_ascii.c
+++ b/cbits/is_ascii.c
@@ -47,8 +47,8 @@ const size_t _hs_text_is_ascii(const uint8_t *src0, const uint8_t *srcend){
 }
 
 /*
-  _hs_text_is_ascii_wrapper is a helper for calling _hs_text_is_ascii on Texts.
+  _hs_text_is_ascii_offset is a helper for calling _hs_text_is_ascii on Texts.
 */
-const size_t _hs_text_is_ascii_2(const uint8_t *arr, size_t off, size_t len){
+const size_t _hs_text_is_ascii_offset(const uint8_t *arr, size_t off, size_t len){
     return _hs_text_is_ascii(arr + off, arr + off + len);
 }
--- a/cbits/is_ascii.c
+++ b/cbits/is_ascii.c
@@ -46,7 +46,9 @@ const size_t _hs_text_is_ascii(const uint8_t *src0, const uint8_t *srcend){
   return src - src0;
 }
 
-// TODO wrapper for using with ByteArray#
+/*
+  _hs_text_is_ascii_wrapper is a helper for calling _hs_text_is_ascii on Texts.
+*/
 const size_t _hs_text_is_ascii_2(const uint8_t *arr, size_t off, size_t len){
     return _hs_text_is_ascii(arr + off, arr + off + len);
 }
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1193,7 +1193,7 @@ minimum t = S.minimum (stream t)
 --
 -- prop> isAscii t == all (< '\x80') t
 --
--- @since 2.0.2 (?) TODO
+-- @since 2.0.2
 isAscii :: Text -> Bool
 isAscii (Text (A.ByteArray arr) off len) =
     cSizeToInt (c_is_ascii_offset arr (intToCSize off) (intToCSize len)) == len
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -266,7 +266,6 @@ import System.IO.Unsafe (unsafePerformIO)
 #endif
 
 import Foreign.Ptr ( plusPtr )
-import System.IO.Unsafe (unsafeDupablePerformIO)
 
 -- $setup
 -- >>> :set -package transformers
@@ -1182,18 +1181,16 @@ minimum t = S.minimum (stream t)
 
 -- TODO
 isAscii :: Text -> Bool
-isAscii (Text (A.ByteArray arr) off len) =
-    let arrPtr = Exts.Ptr (Exts.byteArrayContents# arr)
-        check = c_is_ascii (arrPtr `plusPtr` off) (arrPtr `plusPtr` len)
-    in  cSizeToInt (unsafeDupablePerformIO check) == len
+isAscii (Text (A.ByteArray arr) (Exts.I# off#) (Exts.I# len#)) =
+    cSizeToInt (c_is_ascii_2 arr off# len#) == (Exts.I# len#)
 {-# INLINE isAscii #-}
 
 cSizeToInt :: CSize -> Int
 cSizeToInt = P.fromIntegral
 {-# INLINE cSizeToInt #-}
 
-foreign import ccall unsafe "_hs_text_is_ascii" c_is_ascii
-    :: Exts.Ptr Word8 -> Exts.Ptr Word8 -> IO CSize
+foreign import ccall unsafe "_hs_text_is_ascii_2" c_is_ascii_2
+    :: ByteArray# -> Exts.Int# -> Exts.Int# -> CSize
 
 -- -----------------------------------------------------------------------------
 -- * Building 'Text's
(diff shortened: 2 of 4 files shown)


raehik:
I am almost certain my FFI code is buggy. I don't well understand pinned/unpinned memory. But I'm unable to test on my machine (build issues).

src/Data/Text.hs
```diff
@@ -1176,6 +1180,21 @@ minimum :: HasCallStack => Text -> Char
 minimum t = S.minimum (stream t)
 {-# INLINE minimum #-}
 
+-- TODO
+isAscii :: Text -> Bool
+isAscii (Text (A.ByteArray arr) off len) =
+    let arrPtr = Exts.Ptr (Exts.byteArrayContents# arr)
```
phadej, line 1186: This is unsafe/wrong. `Text`'s bytearray is not pinned, so it may move after you take a pointer. `ccall unsafe` is not enough.

You need to import `_hs_text_is_ascii` as taking `ByteArray#` argument (grep for `memchr`). Then there won't be a race condition.

raehik: I thought that was wrong! Thanks very much for the help, I don't understand memory pinning & had hoped to get poked in the right place. So the FFI resolves `ByteArray#` and `Ptr Word8` to the same C type, but they do different things. Is this hidden in the compiler somewhere? *Edit: I failed to read a line in GHC.Exts about `UnliftedFFITypes`! There's even a table about usage safety! Awesome.*

If you don't mind, another thing I was couldn't figure out: what is the difference between an `IO a` and a "pure" `a` foreign import? (Either way it's the same C return type, right? just info for the compiler?)

phadej: If the function is pure like `sin` then you don't need `IO`.

In

```haskell
foreign import ccall unsafe "_hs_text_is_ascii" c_is_ascii
    :: Ptr Word8 -> Ptr Word8 -> IO CSize
```

The C function reads memory, so that is not pure. (Another thread can write into the same memory range e.g.)

```haskell
foreign import ccall unsafe "_hs_text_is_ascii_2" c_is_ascii_2
    :: ByteArray# -> CSize -> IO CSize
```

is OTOH "pure" as `ByteArray#` is passed in "by value". That's why `IO` was dropped recently (See https://github.com/haskell/text/pull/482)

src/Data/Text.hs
```diff
@@ -1176,6 +1179,19 @@ minimum :: HasCallStack => Text -> Char
...
+isAscii (Text (A.ByteArray arr) (Exts.I# off#) (Exts.I# len#)) =
+    cSizeToInt (c_is_ascii_2 arr off# len#) == (Exts.I# len#)
+{-# INLINE isAscii #-}
+
+cSizeToInt :: CSize -> Int
+cSizeToInt = P.fromIntegral
+{-# INLINE cSizeToInt #-}
+
+foreign import ccall unsafe "_hs_text_is_ascii_2" c_is_ascii_2
+    :: ByteArray# -> Exts.Int# -> Exts.Int# -> CSize
```
phadej, line 1193: you don't need to use `Int#` here. `CSize` works fine (and is actually more correct).

src/Data/Text.hs
```diff
@@ -264,6 +265,8 @@ import Data.Text.Foreign (asForeignPtr)
 import System.IO.Unsafe (unsafePerformIO)
 #endif
 
+import Foreign.Ptr ( plusPtr )
```
phadej, line 268: unused import?

src/Data/Text.hs
raehik: I see! I've done your 2nd example without the `IO`, which calls the main function with pointers to the beginning and end of the data calculated (the same pattern is used in some `Data.Text.Array` foreign calls):

```haskell
isAscii (Text (A.ByteArray arr) (Exts.I# off#) (Exts.I# len#)) =
    cSizeToInt (c_is_ascii_2 arr off# len#) == (Exts.I# len#)
foreign import ccall unsafe "_hs_text_is_ascii_2" c_is_ascii_2
    :: ByteArray# -> Exts.Int# -> Exts.Int# -> CSize
```

src/Data/Text.hs
raehik: From previous definition, removed

phadej (commented):
... and you probably want to add tests.

src/Data/Text.hs
raehik: OK, thanks

raehik:
Fixed my build issues, will do. Thanks a ton for the help with the fiddly bits.

raehik:
Fixed an error in the C code (end of text data is at `arr+off+len`, not `arr+len`), added lazy version, added property tests for both. Ready for review/squish/merge.

raehik:
Something to note: `Data.Text.isAscii` clashes with `Data.Char.isAscii`. There were already a couple of clashing functions (found  `toUpper`, `toLower` on a glance), so I wager that it shouldn't be considered an issue.

cbits/is_ascii.c
```diff
@@ -45,3 +45,10 @@ const size_t _hs_text_is_ascii(const uint8_t *src0, const uint8_t *srcend){
 
   return src - src0;
 }
+
+/*
+  _hs_text_is_ascii_wrapper is a helper for calling _hs_text_is_ascii on Texts.
+*/
+const size_t _hs_text_is_ascii_2(const uint8_t *arr, size_t off, size_t len){
```
Bodigrim, line 52:
```suggestion
const size_t _hs_text_is_ascii_offset(const uint8_t *arr, size_t off, size_t len){
```

raehik: thanks, didn't know what to call this

Lysxia (commented):
That looks good to me!

Another pair of eyes (@phadej?) to go over the FFI bits again would be nice. 
There's the `@since` TODOs, I think this can make it into 2.0.2.

Boarders:
Looks good to me, would be nice to have a test for the edge cases to make sure the offset is working as expected i.e. make sure this fails on text which fails to be ascii on the boundary with non-zero offset (looks correct, but it is useful to have such tests regardless).

src/Data/Text.hs
```diff
@@ -1176,6 +1177,35 @@ minimum :: HasCallStack => Text -> Char
...
+--
+-- >>> isAscii "abc\NUL"
+-- True
+--
+-- >>> isAscii "abcd€"
+-- False
+--
+-- prop> isAscii t == all (< '\x80') t
+--
+-- @since 2.0.2 (?) TODO
```
Bodigrim, line 1196: Yeah, let's commit to 2.0.2 here and below.

raehik: done

raehik:
> Looks good to me, would be nice to have a test for the edge cases to make sure the offset is working as expected i.e. make sure this fails on text which fails to be ascii on the boundary with non-zero offset (looks correct, but it is useful to have such tests regardless).

OK. I added a single test for that:

```haskell
isAscii_border :: IO ()
isAscii_border = do
    -- ASCII prefix ends at position 3 (from 0)
    let text  = T.pack "123一二三"
        text' = case text of T.Text arr off _len -> T.Text arr (off+1) 3
    assertBool "UTF-8 string with ASCII prefix ending at last position incorrectly detected as ASCII" $ not $ T.isAscii text'
```

We're really testing the behaviour of `_hs_text_is_ascii` rather than `isAscii` or the C wrapper, which do very little (a non-zero offset means very little to the C call). What do you think @Boarders ?

tests/Tests/Properties/Folds.hs
```diff
@@ -190,6 +194,13 @@ tl_unfoldrN n m   = (L.take i . L.unfoldr (unf j)) `eq`
     where i = fromIntegral (n :: Word16)
           j = fromIntegral (m :: Word16)
 
+isAscii_border :: IO ()
+isAscii_border = do
+    -- ASCII prefix ends at position 3 (from 0)
+    let text  = T.pack "123一二三"
+        text' = case text of T.Text arr off _len -> T.Text arr (off+1) 3
+    assertBool "UTF-8 string with ASCII prefix ending at last position incorrectly detected as ASCII" $ not $ T.isAscii text'
```
Lysxia, line 202: Instead of pattern-matching and constructing an invalid `Text`, you can use `T.drop` to safely get a non-zero offset. And then the ascii prefix should be longer than the byte-length after `T.drop` to properly test against accidentally ignoring the offset.

Lysxia (approved):

tests/Tests/Properties/Folds.hs
raehik: thanks, made those changes

src/Data/Text/Lazy.hs
```diff
@@ -870,6 +871,27 @@ minimum :: HasCallStack => Text -> Char
...
+--
+-- >>> isAscii "abcd€"
+-- False
+--
+-- prop> isAscii t == all (< '\x80') t
+--
+-- @since 2.0.2
+isAscii :: Text -> Bool
+isAscii Empty = True
+isAscii (Chunk c cs) = T.isAscii c && isAscii cs
```
Bodigrim, line 893: Please use `foldrChunks`.

tests/Tests/Properties/Folds.hs
```diff
@@ -11,14 +11,16 @@ import Control.Arrow (second)
...
+import Test.Tasty.HUnit (testCase, assertFailure, assertBool)
 import Test.Tasty.QuickCheck (testProperty, Small(..), (===), applyFun, applyFun2)
 import Tests.QuickCheckUtils
 import qualified Data.List as L
 import qualified Data.Text as T
 import qualified Data.Text.Internal.Fusion as S
 import qualified Data.Text.Internal.Fusion.Common as S
 import qualified Data.Text.Lazy as TL
+import qualified Data.Char as Char
+import qualified Data.Text.Internal as T
```
Bodigrim, line 23: Redundant import.

Bodigrim:
Thanks!

PR #499, opened 2023-01-28
add safe decodeASCII'

raehik:
Useful function for serialization libs. Currently you have to do something like [this](https://github.com/raehik/binrep/blob/06a609336c25aadd02948aedb2c6f05319050bdb/src/Binrep/Type/Text/Encoding/Ascii.hs#L26-L38), which is nasty and weird for something that isn't really inherently so.

I find the decoding setup to be a little messy and this makes no attempt to improve on that. I wouldn't mind assisting on that work too.

### DIFF
--- a/src/Data/Text/Encoding.hs
+++ b/src/Data/Text/Encoding.hs
@@ -31,6 +31,7 @@ module Data.Text.Encoding
     , decodeASCIIPrefix
     , decodeUtf8Lenient
     , decodeUtf8'
+    , decodeASCII'
 
     -- *** Controllable error handling
     , decodeUtf8With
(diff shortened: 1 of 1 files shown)


raehik:
`decodeAsciiPrefix` in #448 would also resolve this, I doubt anyone minds about a little extra allocation for the fail case.

Lysxia:
And the extra allocation in #448 would get optimized away if you just check the suffix for emptiness.

raehik:
With #448 merged, I'll rebase this and see what's left -- perhaps nothing, or perhaps a convenience definition if it seems sensible.

Lysxia:
Yeah, I think a `Maybe` or `Either` variant of `decodeASCII` makes sense even if it's a simple wrapper around `decodeASCIIPrefix`, because it has a more natural-looking type.

raehik:
Thanks to @Lysxia 's fantastic work on decoding there's not much here now:

```haskell
decodeASCII' :: ByteString -> Maybe Text
decodeASCII' bs =
  let (prefix, suffix) = decodeASCIIPrefix bs in
  if B.empty suffix then Just t else Nothing
{-# INLINE decodeASCII' #-}
```

What do people think? It's perhaps more convenient, but gives less information than `decodeASCIIPrefix`.

src/Data/Text/Encoding.hs
```diff
@@ -177,6 +178,21 @@ asciiPrefixLength bs = unsafeDupablePerformIO $ withBS bs $ \ fp len ->
   unsafeWithForeignPtr fp $ \src -> do
     fromIntegral <$> c_is_ascii src (src `plusPtr` len)
 
+-- | Decode a 'ByteString' containing 7-bit ASCII encoded text.
+--
+-- This is a total function which returns either the 'ByteString' converted to a
+-- 'Text' containing ASCII text, or 'Nothing'.
+--
+-- Use 'decodeASCIIPrefix'' to retain the longest ASCII prefix for an invalid
```
Lysxia, line 186:
```suggestion
-- Use 'decodeASCIIPrefix' to retain the longest ASCII prefix for an invalid
```

Lysxia (approved):

raehik:
pushed fix `B.empty -> B.null` lol and re-squashed

Bodigrim (approved):

Issue #472, opened 2022-09-29
Utf8 IO using `ShortByteString`


oberblastmeister:
I'm thinking we should call it `Text.IO.Utf8`. I made a [pr](https://github.com/haskell/bytestring/pull/547) in bytestring which should be used for this.

PR #503, opened 2023-02-10
Add Text.IO.Utf8 module

oberblastmeister:
Solves #472


Lysxia:
The title of issue #503 mentions using `ShortByteString` for this. What's the status of that?

If I understand correctly, this is about variants of `Data.Text.IO.readFile`, etc. that don't depend on the locale and which may avoid a copy compared to `fmap decodeUtf8 . Data.ByteString.readFile` (although the current implementation doesn't do that). Any other benefit I'm missing? In any case that sounds reasonable to me.

Wouldn't it be better to add to `Data.Text.IO` instead of creating yet another module?

@haskell/text any objections?

oberblastmeister:
Yeah, the intention is to  make it faster in the future once I finish https://github.com/haskell/bytestring/pull/547.

I generally like using qualified imports instead of adding suffixes to everything, so that's why I put it in a separate module. I'm fine with putting it in `Data.Text.IO` also.

Bodigrim:
I'd mildly prefer to fold it into `Data.Text.IO`, but I'm fine either way.

Shall we provide the same set of functions for lazy `Text` as well?

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,44 @@
...
+decodeUtf8IO = evaluate . decodeUtf8
+
+encodeUtf8IO :: Text -> IO ByteString
+encodeUtf8IO = evaluate . encodeUtf8
+
+-- | The 'readFile' function reads a file and returns the contents of
+-- the file as a string.  The entire file is read strictly, as with
+-- 'getContents'.
+readFile :: FilePath -> IO Text
+readFile = decodeUtf8IO <=< B.readFile
```
parsonsmatt, line 35: This operation is partial in the event that the given file *isn't* encoded in UTF-8. I'd personally prefer the signature of `readFile` to return an `Either` with a decode error if that happens. However, I acknowledge that that's contrary to the rest of the API of this package - the convention would be to define `readFile' :: FilePath -> IO (Either DecodeError Text)`.

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,44 @@
...
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+
+decodeUtf8IO :: ByteString -> IO Text
+decodeUtf8IO = evaluate . decodeUtf8
+
+encodeUtf8IO :: Text -> IO ByteString
+encodeUtf8IO = evaluate . encodeUtf8
+
+-- | The 'readFile' function reads a file and returns the contents of
+-- the file as a string.  The entire file is read strictly, as with
```
parsonsmatt, line 32:
```suggestion
-- the file as a strict 'Text'.  The entire file is read strictly, as with
```

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,44 @@
...
+import Data.ByteString (ByteString)
+import qualified Data.ByteString as B
+import Data.Text (Text)
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+
+decodeUtf8IO :: ByteString -> IO Text
+decodeUtf8IO = evaluate . decodeUtf8
+
+encodeUtf8IO :: Text -> IO ByteString
+encodeUtf8IO = evaluate . encodeUtf8
```
parsonsmatt, line 29: I'm curious if this is necessary. Is it possible for `encodeUtf8` to throw an exception? If so, is it possible for throwing at `evaluate` to differ from throwing at `B.writeFile` or `B.appendFile`?

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,44 @@
...
+import Prelude hiding (readFile, writeFile, appendFile)
+import Control.Exception (evaluate)
+import Control.Monad ((<=<))
+import Data.ByteString (ByteString)
+import qualified Data.ByteString as B
+import Data.Text (Text)
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+
+decodeUtf8IO :: ByteString -> IO Text
+decodeUtf8IO = evaluate . decodeUtf8
```
parsonsmatt, line 26: This use of `evaluate` makes sense to me - if `decodeUtf8` throws an error, then we want that to blow up at the `readFile` call-site and not the evaluation site of the returned `Text`.

I feel like this signature would benefit from a `HasCallStack` constraint, since it can throw (along with `readFile` below).

parsonsmatt (commented):
This change looks overall good - especially if we can get a zero-copy via `ShortByteString`!

Is there any intent to deprecate the existing `readFile`, possibly pointing it to a named function `readFileWithSystemLocale`? That would describe the most common footgun and allow people to migrate either to the existing behavior, or to this new behavior.

src/Data/Text/IO/Utf8.hs
oberblastmeister: Yeah I don't think this is necessary

src/Data/Text/IO/Utf8.hs
oberblastmeister: As you said, this doesn't match the convention of the rest of the package. Also, I generally don't like putting `Either` inside of `IO`, because IO can throw other errors too.

src/Data/Text/IO/Utf8.hs
oberblastmeister: I don't think adding it will do anything, because `decodeUtf8` doesn't have `HasCallStack`

oberblastmeister:
How should I group the documentation in the `Data.Text.IO` module?

Lysxia (commented):
When it was just the three file functions it seemed odd to have a dedicated module, but now that I see it's the whole `Text.IO` API that's affected and being duplicated, it makes more sense to have a new module `Data.Text.IO.Utf8`. That way, users who actually don't want to be locale dependent can just do a search-replace on the module name.

oberblastmeister:
@Lysxia @Bodigrim Is there anything else I need to do here?

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,104 @@
...
+
+-- | Write a string to the end of a file.
+appendFile :: FilePath -> Text -> IO ()
+appendFile fp = B.appendFile fp . encodeUtf8
+
+-- | Read the remaining contents of a 'Handle' as a string.  The
+-- 'Handle' is closed once the contents have been read, or if an
+-- exception is thrown.
+--
+-- Internally, this function reads a chunk at a time from the
```
Bodigrim, line 61: I'm not sure this and below is worth retaining: after all this behaviour depends on `bytestring` package, not on us.

src/Data/Text/IO/Utf8.hs
```diff
@@ -0,0 +1,104 @@
+-- |
+-- Module      : Data.Text.IO.Utf8
+-- Copyright   : (c) 2009, 2010 Bryan O'Sullivan,
+--               (c) 2009 Simon Marlow
```
Bodigrim, line 4: It feels weird to copy a decade-old copyright into new files. It seems the content is all new?.. Unless I'm missing something, you can probably omit the field entirely or put your own name if you wish.

Bodigrim (commented):
Overall looks good to me, only couple remarks.

Bodigrim (approved):
LGTM. Any chance to write some tests?

oberblastmeister:
It doesn't seem like Data.Text.IO has any tests? Also, the code really just builds on stuff from bytestring

Bodigrim:
There are some basic IO tests in a very strange location:
https://github.com/haskell/text/blob/ff4af4c2719aa2529bb03450ece3a3f0dc30c95f/tests/Tests/Properties/LowLevel.hs#L141-L145

oberblastmeister:
@Bodigrim I added tests

Lysxia:
There's a danger with the lack of newline conversion. Windows users are going to be very confused when they read their files on their system and get an unexpected number of characters. It's useful to distinguish two issues here:

1. We can read UTF-8 text faster because it's the encoding used internally in Text
2. Often we just know that a file is in UTF-8, so having a shorthand for that case would be useful

We can resolve (1) without changing the API, by adding a special case in the existing IO functions.
But (2) turns out to not be accurate because of newline conversion. UTF-8 still leaves open the question of how to encode newlines. I think that, either way, `Utf8.readFile` and the rest would be a footgun because half of the users will assume it works the other way.

oberblastmeister:
Can't we just document that it doesn't convert newlines? The user can convert newlines explicitly if they want, and we can add functions for that.

How would we add a special case for the existing functions?

Lysxia:
I don't think those functions are really that conventional that they pass the Fairbairn threshold. Even if we document  the behavior, it's really not obvious what specific circumstances warrant it. The standard encoding is set by the platform and the locale. As far as I can tell, it's only legitimate to ignore that standard when you downloaded a file from a Unix system, or you're talking to another local application which is ignoring the standard. IMO The niche bit of convenience of including `Utf8.readFile` in text is not worth the risk that it gets used by default for the wrong reasons, unnecessarily crippling portability. 

> How would we add a special case for the existing functions?

The encoding is determined by the `Handle`, [in the field `haCodec`](https://hackage.haskell.org/package/base-4.18.0.0/docs/src/GHC.IO.Handle.Types.html#Handle__). You can compare its value with  `utf8` and do something different if they're equal.

Bodigrim:
In the modern environment locale-dependent IO is a larger risk than reading UTF-8 by default. You don't really want to lose all data just because someone accidentally changed system locale to ASCII.

E. g., even [GHC itself always expects UTF-8 whatever locale](https://downloads.haskell.org/ghc/latest/docs/users_guide/separate_compilation.html?highlight=utf#haskell-source-files). 

We actually do warn about this issue already:
https://github.com/haskell/text/blob/ff4af4c2719aa2529bb03450ece3a3f0dc30c95f/src/Data/Text/IO.hs#L70-L78

See also https://www.snoyman.com/blog/2016/12/beware-of-readfile/

----

I've defined this very set of functions in private projects more than once, so I'm keen to have them available from `text` directly. They are not replacing existing ones, anyone wishing to be locale-dependent can continue to do so.

Bodigrim:
FWIW `bytestring` has been ignoring locale-dependent line-ending conversion for decades now.

Lysxia:
The blogpost makes a good point. Maybe I underestimated how many system do assume UTF-8. Consider myself overruled. Do mention the lack of newline conversion then.

It makes sense for bytestring to ignore all of that because it's not necessarily dealing with text.

Lysxia (approved):

Bodigrim:
Thanks, @oberblastmeister!

PR #521, opened 2023-06-06
haiku: use libstdc++

jessicah:


jessicah:
Will this get backported to earlier releases as necessary as well, or should I make other PRs for those?

Bodigrim:
There are no plans for backports, we'll just cut `text-2.1.0` at some point in future, probably in time for GHC 9.8.

Bodigrim (approved):

jessicah:
@Bodigrim as a dependency of things like cabal, it'd be helpful to have sooner rather than later, as I need to manually patch, and repackage tarballs for cabal bootstrap, and given it's such a minor change, 2.0.3 seems satisfactory being a non-abi change? E.g. I'm still working with GHC-9.0 presently.

Bodigrim:
For Cabal bootstrap - and especially if you use GHC 9.0 - you should be fine with a vanilla `text-1.2.5.0.tar.gz`.

CI fails because Cabal cannot recognize a new architecture. Instead of adding a clause `elif os(haiku)` you can just act on TODO below (and regenerate `haskell-ci`, upgrading Cabal used by CI jobs to 3.10, if needed).

Lysxia (approved):

PR #534, opened 2023-07-28
Make Lift Text work under RebindableSyntax

### LINKED ISSUE #533: The new `lift` implementation in text-2.0.1 breaks with `RebindableSyntax`
Since text-2.0.1, `lift` implementation for `Text` has changed to, essentially

```haskell
unpackCStringLen# $(bytes) $(lenInt)
```

https://github.com/haskell/text/blob/bc6f4031d6da74bcf97635ac01ac1540e5285672/src/Data/Text.hs#L419-L422

where `lenInt` is an `Int`. The problem here is that `Int` gets lifted into a numeric literal, which gets desugared as `fromInteger <lenInt>`. When using `RebindableSyntax` and there's a non-standard `fromInteger` in scope, this breaks `Text`'s `lift`.

Here's a small repro: https://github.com/lierdakil/text-2.0.1-lift-repro

The only way I could figure a workaround is using an unlifted `Int#` and the primitive constructor. Not too bad, but depends on `GHC.Exts` or somesuch.

Under this suggestion, the relevant code would look something like this (from the top of my head):

```haskell
    let !lenInt@(I# lenIntPrim) = P.fromIntegral len
    TH.appE (TH.appE (TH.varE 'unpackCStringLen#) (TH.litE . TH.bytesPrimL $ TH.mkBytes ptr 0 (P.fromIntegral lenInt))) (conE 'I# `appE` TH.lift lenIntPrim)
```

Lysxia:
The Lift Int instance produces a polymorphic literal, which desugars to fromIntegral under RebindableSyntax. Fixing that instance in base may be too much of a breaking change if a lot of code relies on the accidental polymorphism. We can at least fix Lift Text by not relying on Lift Int.

Fixes #533

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -418,8 +418,10 @@ instance TH.Lift Text where
 #if MIN_VERSION_template_haskell(2,16,0)
   lift txt = do
     let (ptr, len) = unsafePerformIO $ asForeignPtr txt
-    let lenInt = P.fromIntegral len
-    TH.appE (TH.appE (TH.varE 'unpackCStringLen#) (TH.litE . TH.bytesPrimL $ TH.mkBytes ptr 0 lenInt)) (TH.lift lenInt)
+        bytesQ = TH.litE . TH.bytesPrimL $ TH.mkBytes ptr 0 (P.fromIntegral len)
+        lenQ = liftInt (P.fromIntegral len)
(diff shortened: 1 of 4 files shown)


Bodigrim (approved):

Bodigrim:
Great!

PR #548, opened 2023-12-04
Implement measure_off in pure Haskell

hsyl20:
Implement `measure_off` in pure Haskell for use with the JS backend

### DIFF
--- a/src/Data/Text/Internal/Measure.hs
+++ b/src/Data/Text/Internal/Measure.hs
@@ -1,0 +1,53 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+
+#if defined(PURE_HASKELL)
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE MultiWayIf #-}
+#endif
+
+#if !defined(PURE_HASKELL)
+{-# LANGUAGE UnliftedFFITypes #-}
+#endif
+
+{-# OPTIONS_HADDOCK not-home #-}
+
+-- | Implements 'measure_off', using efficient C routines by default.
+module Data.Text.Internal.Measure
+  ( measure_off
+  )
+where
+
+import GHC.Exts
+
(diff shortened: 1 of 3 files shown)


src/Data/Text/Internal/Measure.hs
```diff
@@ -0,0 +1,50 @@
...
+    go !cc !i
+      -- return the number of bytes for the first cnt codepoints,
+      | cc == cnt = fromIntegral i
+      -- return negated number of codepoints if there are less than cnt
+      | i >= len  = negate (fromIntegral cc)
+      | otherwise =
+          let !(I# o) = fromIntegral (off+i)
+              !b = word2Int# (word8ToWord# (indexWord8Array# ba o))
+              -- branchless increment for multi-byte codepoints
+              !i' = i + 1 + fromIntegral (I# ((b >=# 0xC0#) +# (b >=# 0xE0#) +# (b >=# 0xF0#)))
```
Bodigrim, line 40: Please use [`utf8LengthByLeader`](https://hackage.haskell.org/package/text-2.1/docs/Data-Text-Internal-Encoding-Utf8.html#v:utf8LengthByLeader).

src/Data/Text/Internal/Measure.hs
```diff
@@ -0,0 +1,50 @@
...
+
+-- | Implements 'measure_off, using efficient C routines by default.
+module Data.Text.Internal.Measure
+  ( measure_off
+  )
+where
+
+import GHC.Exts
+import GHC.Word
+import Data.Text.Internal.Encoding.Utf8 (utf8LengthByLeader)
```
Bodigrim, line 23: Please put this and `import Data.Word` under `#if defined(PURE_HASKELL)`, otherwise GHC reports a redundant import.

hsyl20: Good catch, done.

Lysxia (approved):

PR #551, opened 2024-01-11
Optimize the implementation of Data.Text.concat

### LINKED ISSUE #550: concat can be faster
I recently wanted a `reverseConcat`, which led me to the current implementation of `concat` and the realization that it is not very efficient:
https://github.com/haskell/text/blob/e84c7a35bba49cd930a2968791636f907bcdb5f0/src/Data/Text.hs#L1033-L1046

* Allocates a new list of `Text`s (lazily, but it still has a cost) because of the filter
* Allocates a list of boxed int lengths to sum them, since `sumP` does not fuse

Here is an alternate straightforward implementation:
```hs
concat :: [T.Text] -> T.Text
concat ts0 = case ts0 of
  [] -> T.empty
  [t] -> t
  _ | len == 0 -> T.empty
    | otherwise -> T.Text arr 0 len
  where
    flen acc (T.Text _ _ l)
      | acc' > 0 || l == 0 = acc'
      | otherwise = concatOverflowError
      where
        acc' = acc + l
    len = foldl' flen 0 ts0
    arr = A.run $ do
      marr <- A.new len
      let loop !i [] = pure marr
          loop i (T.Text a o l : ts) = A.copyI l marr i a o *> loop (i+l) ts
      loop 0 ts0
```

This not _exactly_ the same, since the current implementation performs a case match on the list after filtering out null `Text`s. But I doubt such a preemptive step helps.

Benchmarks on GHC 9.6.3 with -O, concating a list of all `Char`s:

```
    concat:     OK
      54.8 ms ± 3.1 ms, 174 MB allocated,  52 MB copied, 325 MB peak memory
    alt concat: OK
      13.1 ms ± 901 μs, 4.2 MB allocated, 773 B  copied, 325 MB peak memory
    ==:         OK
```

<details>
<summary>Benchmark file</summary>

```hs
{-# LANGUAGE BangPatterns #-}
import Prelude hiding (concat)
import Data.Foldable (foldl')
import Test.Tasty.Bench
import Test.Tasty.HUnit

import qualified Data.Text.Internal as T
import qualified Data.Text.Array as A

main :: IO ()
main = defaultMain
  [ env (pure xs_) $ \xs -> bgroup ""
    [ bench "concat" $ whnf T.concat xs
    , bench "alt concat" $ whnf concat xs
    , testCase "==" $ concat xs @?= T.concat xs
    ]
  ]
  where
    xs_ = map T.singleton [minBound .. maxBound]

concat :: [T.Text] -> T.Text
concat ts0 = case ts0 of
  [] -> T.empty
  [t] -> t
  _ | len == 0 -> T.empty
    | otherwise -> T.Text arr 0 len
  where
    flen acc (T.Text _ _ l)
      | acc' > 0 || l == 0 = acc'
      | otherwise = concatOverflowError
      where
        acc' = acc + l
    len = foldl' flen 0 ts0
    arr = A.run $ do
      marr <- A.new len
      let loop !i [] = pure marr
          loop i (T.Text a o l : ts) = A.copyI l marr i a o *> loop (i+l) ts
      loop 0 ts0

concatOverflowError :: a
concatOverflowError = error "Data.Text.concat: size overflow"
```
</details>

meooow25:
Avoids allocating intermediate lists. Does not make any change to the algorithm.

Fixes #550.

As mentioned in https://github.com/haskell/text/issues/550#issuecomment-1880244020, `concat` can alternately be implemented using a strategy which makes one pass over the list, filling and growing a buffer. This PR only optimizes the existing implementation and _does not_ change to this strategy.

Output of `cabal run text-benchmarks -- -p '$4 == "concat" && $5 == "Text"' +RTS -T ` with GHC 9.6.3:

<details>
<summary>Before:</summary>

```
All
  Pure
    tiny
      concat
        Text: OK
          15.5 ns ± 1.3 ns,  79 B  allocated,   0 B  copied,  35 MB peak memory
    ascii-small
      concat
        Text: OK
          24.4 μs ± 1.3 μs, 147 KB allocated,  17 B  copied,  62 MB peak memory
    ascii
      concat
        Text: OK
          41.1 ms ± 3.5 ms, 136 MB allocated,  15 MB copied, 511 MB peak memory
    english
      concat
        Text: OK
          2.17 ms ± 213 μs, 8.5 MB allocated,  14 KB copied, 511 MB peak memory
    russian
      concat
        Text: OK
          2.53 μs ± 194 ns,  18 KB allocated,   0 B  copied, 511 MB peak memory
    japanese
      concat
        Text: OK
          6.00 μs ± 251 ns,  39 KB allocated,   2 B  copied, 511 MB peak memory
```
</details>

After:
```
  Pure
    tiny
      concat
        Text: OK
          4.67 ns ± 346 ps,  23 B  allocated,   0 B  copied,  35 MB peak memory, 69% less than baseline
    ascii-small
      concat
        Text: OK
          12.1 μs ± 964 ns,  66 KB allocated,   0 B  copied,  73 MB peak memory, 50% less than baseline
    ascii
      concat
        Text: OK
          16.3 ms ± 726 μs,  57 MB allocated, 951 B  copied, 498 MB peak memory, 60% less than baseline
    english
      concat
        Text: OK
          874  μs ±  73 μs, 3.8 MB allocated,  25 B  copied, 498 MB peak memory, 59% less than baseline
    russian
      concat
        Text: OK
          1.65 μs ± 159 ns,  12 KB allocated,   0 B  copied, 498 MB peak memory, 34% less than baseline
    japanese
      concat
        Text: OK
          3.22 μs ± 157 ns,  12 KB allocated,   0 B  copied, 498 MB peak memory, 46% less than baseline
```

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1032,18 +1032,18 @@ foldr' f z t = S.foldl' (P.flip f) z (reverseStream t)
 
 -- | /O(n)/ Concatenate a list of 'Text's.
 concat :: [Text] -> Text
-concat ts = case ts' of
-              [] -> empty
-              [t] -> t
-              _ -> Text (A.run go) 0 len
+concat ts = case ts of
+    [] -> empty
+    [t] -> t
+    _ | len == 0 -> empty
+      | otherwise -> Text (A.run go) 0 len
   where
-    ts' = L.filter (not . null) ts
-    len = sumP "concat" $ L.map lengthWord8 ts'
+    len = sumP "concat" $ L.map lengthWord8 ts
     go :: ST s (A.MArray s)
     go = do
       arr <- A.new len
       let step i (Text a o l) = A.copyI l arr i a o >> return (i + l)
-      foldM step 0 ts' >> return arr
+      foldM step 0 ts >> return arr
 
 -- | /O(n)/ Map a function over a 'Text' that results in a 'Text', and
 -- concatenate the results.
@@ -2043,12 +2043,12 @@ stripSuffix p@(Text _arr _off plen) t@(Text arr off len)
 
 -- | Add a list of non-negative numbers.  Errors out on overflow.
 sumP :: String -> [Int] -> Int
-sumP fun = go 0
-  where go !a (x:xs)
-            | ax >= 0   = go ax xs
+sumP fun = L.foldl' add 0
+  where add a x
+            | ax >= 0   = ax
             | otherwise = overflowError fun
           where ax = a + x
-        go a  _         = a
+{-# INLINE sumP #-} -- Use foldl' and inline for fusion.
 
 emptyError :: HasCallStack => String -> a
 emptyError fun = P.error $ "Data.Text." ++ fun ++ ": empty input"


src/Data/Text.hs
```diff
@@ -1032,18 +1032,18 @@ foldr' f z t = S.foldl' (P.flip f) z (reverseStream t)
...
   where
-    ts' = L.filter (not . null) ts
-    len = sumP "concat" $ L.map lengthWord8 ts'
+    len = sumP "concat" $ L.map lengthWord8 ts
     go :: ST s (A.MArray s)
     go = do
       arr <- A.new len
       let step i (Text a o l) = A.copyI l arr i a o >> return (i + l)
-      foldM step 0 ts' >> return arr
+      foldM step 0 ts >> return arr
```
meooow25, line 1046: Note: I noticed that the index in this loop `i` stays boxed with `-O`, but not with `-O2`. Since `text` is compiled with `-O2`, it's not a problem, but let me know if you would like this changed to something that is unboxed with `-O`.

Bodigrim (approved):

src/Data/Text.hs
```diff
@@ -2043,12 +2043,12 @@ stripSuffix p@(Text _arr _off plen) t@(Text arr off len)
 
 -- | Add a list of non-negative numbers.  Errors out on overflow.
 sumP :: String -> [Int] -> Int
-sumP fun = go 0
-  where go !a (x:xs)
-            | ax >= 0   = go ax xs
+sumP fun = L.foldl' add 0
```
Lysxia, line 2046: Let's add a comment that `sumP` uses `foldl'` to fuse with the `L.map` in `concat` because I don't see myself remembering this if I look at just the code in a few months.

src/Data/Text.hs
Lysxia: As you say, I'm also curious to know what the optimizer at `-O` is missing but `-O2` for text seems a safe enough assumption.

Lysxia (commented):
Looks pretty good. I left a tiny suggestion.

src/Data/Text.hs
meooow25: Added a comment.

src/Data/Text.hs
meooow25: Yes I wouldn't be able to explain why this needs `O2` (`fspec-constr` is also enough).  
If you want me to tweak this, this is as small a change as `!_ <- foldM step 0 ts`. But I'm okay with leaving this as is if you are.

meooow25 (commented):
Updated.
Also added a note in the commit message about a case I thought of where the behavior changes from the current implementation.

> Note: In case there is one non-empty Text and one or more empty Texts,
the non-empty Text is copied whereas it would previously be returned as
is. The behavior is unchanged for every other case.

Lysxia (approved):

PR #552, opened 2024-01-24
Fix typo

horus:
This PR fixes a typo that can only be seen when the `PURE_HASKELL` flag is enabled.


Bodigrim (approved):

Bodigrim:
This is trivial, but very impactful, so let me merge it straight away. Thanks!

PR #554, opened 2024-01-27
Reenable -fpure-haskell

chreekat:
The other implementation looks identical:

```haskell
reverseNonEmpty (Text (A.ByteArray ba) off len) = runST $ do
    marr@(A.MutableByteArray mba) <- A.new len
    unsafeIOToST $ c_reverse mba ba (fromIntegral off) (fromIntegral len)
    brr <- A.unsafeFreeze marr
    return $ Text brr 0 len
```

### DIFF
--- a/src/Data/Text/Internal/Transformation.hs
+++ b/src/Data/Text/Internal/Transformation.hs
@@ -22,7 +22,6 @@
 
 module Data.Text.Internal.Transformation
   ( mapNonEmpty
-  , reverseNonEmpty
   , toCaseFoldNonEmpty
   , toLowerNonEmpty
   , toUpperNonEmpty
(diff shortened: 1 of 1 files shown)


TeofilC:
Thanks for spotting and fixing this!

Bodigrim:
I'm somewhat out of brain power today: how does removing a bunch of code not break anything but instead fixes something? ;)

chreekat:
Understandable! :D  

It looks like Teo moved reverseNonEmpty from its original location into a new module. I did the same thing, but into a different module. My version also got reimplemented to work with -fpure-haskell, but Teo's did not. Finally, by the mysterious ways of git, it's *my* reverseNonEmpty that is imported into Data.Text.Lazy.

Thus, Teo's Data.Text.Internal.Transformation.reverseNonEmpty is unused within the library itself, and fails to compile with -fpure-haskell. Since it is identical to my function (under -fno-pure-haskell, that is), removing it fixes the problem.

Bodigrim (approved):

Lysxia (approved):

PR #558, opened 2024-02-07
Feature: `initsNE` and `tailsNE`.

### LINKED ISSUE #557: Add initsNE, tailsNE :: Text -> NonEmpty Text
Let's follow the suite of `bytestring`, which provides [`initsNE :: ByteString -> NonEmpty ByteString`](https://hackage.haskell.org/package/bytestring-0.12.0.2/docs/Data-ByteString.html#v:initsNE) and [`tailsNE :: ByteString -> NonEmpty ByteString`](https://hackage.haskell.org/package/bytestring-0.12.0.2/docs/Data-ByteString.html#v:tailsNE). 

This would improve consistency between `bytestring` and `text` and dovetail nicely in the efforts to eliminate `Prelude.head` / `Prelude.tail`, allowing to use `Data.List.NonEmpty.head` in more situations.

For the reference, here is a relevant PR for `bytestring`: https://github.com/haskell/bytestring/pull/557

CC @kindaro.

kindaro:
Resolve #557.

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1534,7 +1534,6 @@ inits t = empty : case t of
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
 initsNE :: Text -> NonEmptyList.NonEmpty Text
--- see Note [Avoid NonEmpty combinators]
 initsNE = P.error "Not implemented."
 
 -- | /O(n)/ Return all final segments of the given 'Text', longest
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -230,7 +230,6 @@ import Control.Monad.ST (ST, runST)
 import qualified Data.Text.Array as A
 import qualified Data.List as L hiding (head, tail)
 import qualified Data.List.NonEmpty as NonEmptyList
-import Data.Maybe (fromMaybe)
 import Data.Binary (Binary(get, put))
 import Data.Monoid (Monoid(..))
 import Data.Semigroup (Semigroup(..))
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -230,6 +230,7 @@ import Control.Monad.ST (ST, runST)
 import qualified Data.Text.Array as A
 import qualified Data.List as L hiding (head, tail)
 import qualified Data.List.NonEmpty as NonEmptyList
+import Data.Maybe (fromMaybe)
 import Data.Binary (Binary(get, put))
 import Data.Monoid (Monoid(..))
 import Data.Semigroup (Semigroup(..))
@@ -1534,7 +1535,7 @@ inits t = empty : case t of
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
 initsNE :: Text -> NonEmptyList.NonEmpty Text
-initsNE = P.error "Not implemented."
+initsNE = fromMaybe P.undefined . NonEmptyList.nonEmpty . inits
 
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
@@ -1545,7 +1546,7 @@ tails t | null t    = [empty]
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
 tailsNE :: Text -> NonEmptyList.NonEmpty Text
-tailsNE = P.error "Not implemented."
+tailsNE = fromMaybe P.undefined . NonEmptyList.nonEmpty . tails
 
 -- $split
 --
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1525,11 +1525,7 @@ group = groupBy (==)
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
 inits :: Text -> [Text]
-inits t = empty : case t of
-  Text arr off len ->
-    let loop i | i >= len = []
-               | otherwise = let !j = i + iter_ t i in Text arr off j : loop j
-    in loop 0
+inits = NonEmptyList.toList . initsNE
 
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1546,7 +1546,7 @@ tails = NonEmptyList.toList . tailsNE
 -- first.
 tailsNE :: Text -> NonEmptyList.NonEmpty Text
 tailsNE t
-  | null t = NonEmptyList.singleton empty
+  | null t = empty NonEmptyList.:| []
   | otherwise = t NonEmptyList.:| tails (unsafeTail t)
 
 -- $split
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1529,6 +1529,8 @@ inits = (NonEmptyList.toList $!) . initsNE
 
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
+--
+-- @since 2.1.2
 initsNE :: Text -> NonEmptyList.NonEmpty Text
 initsNE t = empty NonEmptyList.:| case t of
   Text arr off len ->
@@ -1544,6 +1546,8 @@ tails = (NonEmptyList.toList $!) . tailsNE
 
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
+--
+-- @since 2.1.2
 tailsNE :: Text -> NonEmptyList.NonEmpty Text
 tailsNE t
   | null t = empty NonEmptyList.:| []
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1535,7 +1535,12 @@ inits t = empty : case t of
 -- | /O(n)/ Return all initial segments of the given 'Text', shortest
 -- first.
 initsNE :: Text -> NonEmptyList.NonEmpty Text
-initsNE = fromMaybe P.undefined . NonEmptyList.nonEmpty . inits
+initsNE t = empty NonEmptyList.:| case t of
+  Text arr off len ->
+    let loop i
+          | i >= len = []
+          | otherwise = let !j = i + iter_ t i in Text arr off j : loop j
+    in loop 0
 
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
@@ -1546,7 +1551,9 @@ tails t | null t    = [empty]
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
 tailsNE :: Text -> NonEmptyList.NonEmpty Text
-tailsNE = fromMaybe P.undefined . NonEmptyList.nonEmpty . tails
+tailsNE t
+  | null t = NonEmptyList.singleton empty
+  | otherwise = t NonEmptyList.:| tails (unsafeTail t)
 
 -- $split
 --
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -155,7 +155,9 @@ module Data.Text
     , group
     , groupBy
     , inits
+    , initsNE
     , tails
+    , tailsNE
 
     -- ** Breaking into many substrings
     -- $split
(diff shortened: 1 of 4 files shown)


kindaro:
@Bodigrim   Does this look somewhat like what you had in mind?

kindaro:
I have some questions regarding performance.

Note that there are no benchmarks for `inits` and `tails` in `master` right now. I tried to add them but they eat a lot of memory and did not terminate in 10 minutes. So we are flying blind here.

* If we define `inits` in terms of `initsNE`, should we add the `inline` pragma? I am a cargo cultist when it comes to inlining, I need a consultation of an ordained priest.

* If `tails` and `tailsNE` are mutually recursive, would it not mean that lists and non-empty lists get converted into one another on each iteration? It seems we should not define `inits` in terms of `initsNE`, as in the `bytestring` patch <https://github.com/haskell/bytestring/pull/557>, but rather keep the definition of `inits` as it was.

Bodigrim:
Thanks!

> * If we define `inits` in terms of `initsNE`, should we add the `inline` pragma to `inits` to get rid of this indirection? I am a cargo cultist when it comes to inlining, I need a consultation of an ordained priest.

Historically `text` was keen to put `INLINE` everywhere. This helps to game microbenchmarks, but bad for code bloat and compilation time, so nowadays I'm reluctant to add it. It's too blunt of a hammer.

I think it's fine as is. Maybe add `{-# INLINABLE #-}` and see if it changes anything. 

> * If `tails` and `tailsNE` are mutually recursive (as they are now), would it not mean that lists and non-empty lists get converted into one another on each iteration?

I suspect that `tailsNE` gets inlined into `tails` and `:|` gets replaced by `:`, so there is no ping-pong, but I have not checked Core. @clyring any recollection about it?

src/Data/Text.hs
```diff
@@ -1522,17 +1525,29 @@ group = groupBy (==)
...
+          | i >= len = []
+          | otherwise = let !j = i + iter_ t i in Text arr off j : loop j
     in loop 0
 
 -- | /O(n)/ Return all final segments of the given 'Text', longest
 -- first.
 tails :: Text -> [Text]
-tails t | null t    = [empty]
-        | otherwise = t : tails (unsafeTail t)
+tails = NonEmptyList.toList . tailsNE
```
Bodigrim, line 1543: IIRC it's important to use 
```
notThatLazyToList (a :| as) -> a : as
```
instead of [`NonEmptyList.toList`](https://hackage.haskell.org/package/base-4.19.0.0/docs/src/Data.List.NonEmpty.html#toList) (watch for `~`) so that stuff gets inlined and optimized away.

clyring:
> > * If `tails` and `tailsNE` are mutually recursive (as they are now), would it not mean that lists and non-empty lists get converted into one another on each iteration?
> 
> I suspect that `tailsNE` gets inlined into `tails` and `:|` gets replaced by `:`, so there is no ping-pong, but I have not checked Core. @clyring any recollection about it?

If memory serves, `tailsNE` gets a worker/wrapper split and when its wrapper inlines into `tails` the intermediate `(:|)` constructor is easily optimized away as long as an appropriately strict `nonEmptyListToList` conversion is used. (The `Data.List.NonEmpty.toList` provided by `base` is too lazy for this purpose... and so is `Data.List.NonEmpty.cons` and so on...) This optimization should be very reliable, but manually writing the logic in a function with type `Text -> (# Text, [Text] #)` or similar is another option, if you prefer to be explicit about it but don't want the logic duplicated or partial-ified.

kindaro:
I wish we had a way to check performance, but changing the design of benchmarks is out of scope of this issue.

* Some workflows failed because `singleton` was not exported from `base` until GHC 9. I removed the use of `singleton` so that the code builds with GHC 8.10.7. @Bodigrim would it be affordable to run the workflows again?
* As Matthew says _(or, well, as I understand him)_, we need to make sure the outermost constructor of the output of `initsNE` in the definition of `inits` is forced, and likewise for `tails`. I accomplish this by writing in some `$!` operators. @clyring does this look right to you?

Another question: should I add `since` metadata to documentation? What version should I specify? Right now it seems the version is 2.1. It seems to me 2.1.1 would be fitting.

Bodigrim:
> @Bodigrim would it be affordable to run the workflows again?

Sure, sorry for being slow with it.

> Another question: should I add since metadata to documentation? What version should I specify? Right now it seems the version is 2.1. It seems to me 2.1.1 would be fitting.

We are going to release 2.1.1 very soon (ideally tomorrow), so let's put 2.1.2.

src/Data/Text.hs
kindaro: Strictness was added by writing `(toList $!)` instead of `toList`. Hopefully this will work.

kindaro:
Turns out some workflows cannot handle Unicode in the standard output of the test suite, so I replaced the Unicode ellipsis with three ASCII dots. All workflows should pass now. Also, I added `since` metadata to each of the 4 new functions.

@Bodigrim Let me know if this looks good, so that I can rebase and rewrite the commit history into something nice.

src/Data/Text/Lazy.hs
```diff
@@ -1427,18 +1429,32 @@ groupBy eq (Chunk t ts) = cons x ys : groupBy eq zs
...
+inits = (NE.toList P.$!) . initsNE
+
+-- | /O(n)/ Return all initial segments of the given 'Text',
+-- shortest first.
+--
+-- @since 2.1.2
+initsNE :: Text -> NonEmpty Text
+initsNE = (Empty NE.:|) . inits'
   where inits' Empty        = []
         inits' (Chunk t ts) = L.map (\t' -> Chunk t' Empty) (L.drop 1 (T.inits t))
```
Lysxia, line 1441:
```suggestion
        inits' (Chunk t ts) = L.map (\t' -> Chunk t' Empty) (NE.tail (T.initsNE t))
```

`L.drop` does not inline because it is recursive.

Lysxia:
Rather than finicky benchmarks, in this case it is easy enough to see that the optimized Core looks reasonable (no accidental thunks), by adding the following options at the top of the relevant files:

```
{-# OPTIONS_GHC -ddump-simpl -dsuppress-all -ddump-to-file #-}
```

### Core of `Data.Text`

`inits`

```
initsNE
  = \ t_sdIN ->
      :|
        empty
        (case t_sdIN of { Text bx_dalv bx1_dalw bx2_dalx ->
         letrec {
           $wloop_sdIL
             = \ ww_sdII ->
                 case >=# ww_sdII bx2_dalx of {
                   __DEFAULT ->
                     case indexWord8Array# bx_dalv (+# bx1_dalw ww_sdII) of r#_ibpI
                     { __DEFAULT ->
                     let {
                       c#_sbIo
                         = word2Int# (clz8# (and# (not# (word8ToWord# r#_ibpI)) 255##)) } in
                     let { j_sbIq = +# ww_sdII (xorI# c#_sbIo (<=# c#_sbIo 0#)) } in
                     : (Text bx_dalv bx1_dalw j_sbIq) ($wloop_sdIL j_sbIq)
                     };
                   1# -> []
                 }; } in
         $wloop_sdIL 0#
         })

inits
  = \ x_X7 ->
      : empty
        (case x_X7 of { Text bx_dalv bx1_dalw bx2_dalx ->
         letrec {
           $wloop_sdIL
             = \ ww_sdII ->
                 case >=# ww_sdII bx2_dalx of {
                   __DEFAULT ->
                     case indexWord8Array# bx_dalv (+# bx1_dalw ww_sdII) of r#_ibpI
                     { __DEFAULT ->
                     let {
                       c#_sbIo
                         = word2Int# (clz8# (and# (not# (word8ToWord# r#_ibpI)) 255##)) } in
                     let { j_sbIq = +# ww_sdII (xorI# c#_sbIo (<=# c#_sbIo 0#)) } in
                     : (Text bx_dalv bx1_dalw j_sbIq) ($wloop_sdIL j_sbIq)
                     };
                   1# -> []
                 }; } in
         $wloop_sdIL 0#
         })
```

`tails`

```
Rec {
$wtailsNE
  = \ ww_sdp1 ww1_sdp2 ww2_sdp3 ->
      case <=# ww2_sdp3 0# of {
        __DEFAULT ->
          (# Text ww_sdp1 ww1_sdp2 ww2_sdp3,
             case indexWord8Array# ww_sdp1 ww1_sdp2 of r#_ib5C { __DEFAULT ->
             let {
               c#_ib5D
                 = word2Int# (clz8# (and# (not# (word8ToWord# r#_ib5C)) 255##)) } in
             let { y_ib5E = xorI# c#_ib5D (<=# c#_ib5D 0#) } in
             case $wtailsNE ww_sdp1 (+# ww1_sdp2 y_ib5E) (-# ww2_sdp3 y_ib5E) of
             { (# ww3_sdPk, ww4_sdPl #) ->
             : ww3_sdPk ww4_sdPl
             }
             } #);
        1# -> (# empty, [] #)
      }
end Rec }

tailsNE
  = \ t_sdoZ ->
      case t_sdoZ of { Text ww_sdp1 ww1_sdp2 ww2_sdp3 ->
      case $wtailsNE ww_sdp1 ww1_sdp2 ww2_sdp3 of
      { (# ww3_sdPk, ww4_sdPl #) ->
      :| ww3_sdPk ww4_sdPl
      }
      }

tails
  = \ x_iaLw ->
      case x_iaLw of { Text ww_sdp1 ww1_sdp2 ww2_sdp3 ->
      case $wtailsNE ww_sdp1 ww1_sdp2 ww2_sdp3 of
      { (# ww3_sdPk, ww4_sdPl #) ->
      : ww3_sdPk ww4_sdPl
      }
      }
```

kindaro:
Thank you Li-yao.

> Rather than finicky benchmarks, in this case it is easy enough to see that the optimized Core looks reasonable (no accidental thunks), by adding the following options at the top of the relevant files:

Adding some flags is easy, it is understanding the output, written _(or, rather, generated)_ in a language for which there is no documentation, that is challenging. For now I can only accept your observations on faith since I have no expertise to tell if your reasoning is fair. I committed the change to `initsNE.inits'` you suggested — it looks better in Haskell — but I cannot say if it will be more efficient, or less efficient, or without significant difference in efficiency.

By the way, you can also add `-dsuppress-uniques` so GHC will not needlessly obfuscate names.

I think next time I do something with `text` I should commit the baseline simplifier output and then see how it changes after the Haskell source is modified. This would be curious.

The simplifier output for `Data.Text.Lazy.inits_inits'` is the same between `master` and 0458012c30ea7150c060e9613eb833e424a2cdcf. So, I did not break anything — if something is broken, it was broken before I touched it.

kindaro:
@Bodigrim   Please let me know what the next step is to be.

Bodigrim (approved):
LGTM, thanks!

src/Data/Text/Lazy.hs
```diff
@@ -1427,18 +1429,32 @@ groupBy eq (Chunk t ts) = cons x ys : groupBy eq zs
 -- | /O(n)/ Return all initial segments of the given 'Text',
 -- shortest first.
 inits :: Text -> [Text]
-inits = (Empty :) . inits'
+inits = (NE.toList P.$!) . initsNE
+
+-- | /O(n)/ Return all initial segments of the given 'Text',
```
meooow25, line 1434: The $O(n)$ claim is incorrect, this is $O(n^2)$ due to `L.map (Chunk t) (inits' ts)`.
But this was not introduced in the PR, so maybe it can be fixed separately.

kindaro: Alright, I changed the documentation for `Data.Text.Lazy.inits` and `Data.Text.Lazy.initsNE` to say _O(n²)_.

kindaro: I do not follow your reasoning however.

* `map` will take time linear in the amount of chunks.
* `Chunk t` will take constant time.

So, it seems to me `map (Chunk t)` should be linear. What am I missing?

Lysxia: `map (Chunk t)` is linear, and you do it once per chunk. That makes the complexity quadratic.

More accurately, the complexity is `O(cn)` where `c` is the number of chunks (and `n = Lazy.Text.length input`).

Bodigrim:
@Lysxia @meooow25 any more comments?

meooow25:
I haven't checked the core but otherwise seems okay to me 👍

Lysxia (approved):

Bodigrim:
Thanks @kindaro!

PR #560, opened 2024-02-14
Fix filter/filter rules for Text and lazy Text

### LINKED ISSUE #559: filter/filter rules are incorrect
```hs
main = print $ T.filter (const undefined) $ T.filter (const False) $ T.singleton 'a'
```
Without optimizations, prints
```
""
```
With optimizations,
```
tfilter: Prelude.undefined
CallStack (from HasCallStack):
  undefined, called at tfilter.hs:2:32 in main:Main
```

https://github.com/haskell/text/blob/456783a6471ffc0d7dd2a7c0bb38ce43e073ac1b/src/Data/Text.hs#L1660-L1663
https://github.com/haskell/text/blob/456783a6471ffc0d7dd2a7c0bb38ce43e073ac1b/src/Data/Text/Lazy.hs#L1691-L1694

The problem is that the predicate of the inner filter is not checked first.

The rule in the stream code is, however, correct

https://github.com/haskell/text/blob/456783a6471ffc0d7dd2a7c0bb38ce43e073ac1b/src/Data/Text/Internal/Fusion/Common.hs#L1113-L1116

meooow25:
Fixes #559

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -1686,7 +1686,7 @@ filter p = filter_ text p
 
 {-# RULES
 "TEXT filter/filter -> filter" forall p q t.
-    filter p (filter q t) = filter (\c -> p c && q c) t
+    filter p (filter q t) = filter (\c -> q c && p c) t
 #-}
 
 -- | /O(n+m)/ Find the first instance of @needle@ (which must be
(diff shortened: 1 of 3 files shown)


Lysxia (approved):

meooow25:
Done!

Bodigrim (approved):

Bodigrim:
Thanks!

PR #564, opened 2024-02-21
Upgrade simdutf to 4.0.9

### LINKED ISSUE #563: Can simdutf be updated please?
The version of simdutf in haskell/text doesn't correctly check if the OS works with AVX512 (resulting in SIGILL on AVX512 machines where there's no OS support). Would it be possible to update to a version after https://github.com/simdutf/simdutf/commit/55b107f609f5f63880db650a92861ae84cb10abe which fixes this please?

Lysxia:
Resolves #563


Bodigrim (approved):
I downloaded https://github.com/simdutf/simdutf/releases/download/v4.0.9/singleheader.zip and verified that the contents matches.

PR #565, opened 2024-02-21
Improve precision of Data.Text.Read.rational

### LINKED ISSUE #34: `Data.Text.Read.rational` no more accurate than `Data.Text.Read.double`
The documentation on `double` states that it loses accuracy on some inputs, but using `rational` will fix this. I understood this to mean that if one simply used `rational :: Reader Double`, then one would get completely accurate results, but this doesn't appear to be the case. For instance, given the input string  `"5.781884674518029e-2"`, `double :: Reader Double` and `rational :: Reader Double` both produce an output of `"5.7818846745180286e-2"`, whereas `Prelude.read :: String -> Double` produces `"5.781884674518029e-2"`.

Perhaps the documentation means that one should actually use `rational :: Reader Rational` and then convert to `Double` with `fromRational`? If so, that isn't clear from the documentation.

Bodigrim:
Fixes #34.

### DIFF
--- a/src/Data/Text/Lazy/Read.hs
+++ b/src/Data/Text/Lazy/Read.hs
@@ -134,8 +134,16 @@ signed f = runP (signa (P f))
 -- >rational "3e"    == Right (3.0, "e")
 rational :: Fractional a => Reader a
 {-# SPECIALIZE rational :: Reader Double #-}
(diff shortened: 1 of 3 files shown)


Lysxia (approved):
lgtm

PR #566, opened 2024-02-23
Don't try to use AVX512 without OS support

silby:
AVX512 instructions require OS support as well as CPU support. Bits 5, 6, and 7 of XCR0 should be enabled to run with AVX512 support. This fixes SIGILL on OpenBSD, for example.

[OpenBSD ports commit with identical patch](https://github.com/openbsd/ports/commit/bb29df359f152890503a01cd3157eeb0e9701c9d)

Required alongside the similar fix to simdutf pulled in by #564 for text to work right on OpenBSD when the processor flags AVX512 support.

### DIFF
--- a/cbits/measure_off.c
+++ b/cbits/measure_off.c
@@ -44,12 +44,16 @@
 bool has_avx512_vl_bw() {
 #if (__GNUC__ >= 7 || __GNUC__ == 6 && __GNUC_MINOR__ >= 3) || defined(__clang_major__)
   uint32_t eax = 0, ebx = 0, ecx = 0, edx = 0;
+  uint64_t xcr0;
   __get_cpuid_count(7, 0, &eax, &ebx, &ecx, &edx);
   // https://en.wikipedia.org/wiki/CPUID#EAX=7,_ECX=0:_Extended_Features
+  __asm__("xgetbv\n\t" : "=a" (xcr0) : "c" (0));
   const bool has_avx512_bw = ebx & (1 << 30);
   const bool has_avx512_vl = ebx & (1 << 31);
+  // XCR0 bits 5, 6, and 7
+  const bool avx512_os_enabled = (xcr0 & 0xE0) == 0xE0;
   // printf("cpuid=%d=cpuid\n", has_avx512_bw && has_avx512_vl);
-  return has_avx512_bw && has_avx512_vl;
+  return has_avx512_bw && has_avx512_vl && avx512_os_enabled;
 #else
   return false;
 #endif


cbits/measure_off.c
```diff
@@ -44,12 +44,16 @@
 bool has_avx512_vl_bw() {
 #if (__GNUC__ >= 7 || __GNUC__ == 6 && __GNUC_MINOR__ >= 3) || defined(__clang_major__)
   uint32_t eax = 0, ebx = 0, ecx = 0, edx = 0;
+  uint64_t xcr0;
   __get_cpuid_count(7, 0, &eax, &ebx, &ecx, &edx);
   // https://en.wikipedia.org/wiki/CPUID#EAX=7,_ECX=0:_Extended_Features
+  // __asm__("xgetbv\n\t" : "=a" (xcr0) : "c" (0));
```
Bodigrim, line 50: Is it supposed to be commented out?

silby: Oh you know what, definitely not, looks like I have to go bug someone about fixing the patch in ports also. Will force-push in a second.

Bodigrim:
@silby is it possible to setup a regression CI job? Which OS do not support AVX512?

silby:
Well the only one I know about is OpenBSD, which is where we discovered the problem. ([Whole saga here](https://marc.info/?t=170811304700001&r=1&w=2).) I'm not an expert on really anything involved here I just happened to try to use pandoc on OpenBSD-current on my 11th-gen Intel laptop and banged my head against it until I found enough clues. So I think realistic regression-testing would require specific hardware for the test runner, since it would have to have a processor indicating AVX512 support.

Bodigrim:
@blackgnezdo could you please advise?

blackgnezdo:
Since @silby figured this out and tested the fix, I'm pretty comfortable with suggesting this gets merged.

I don't have any insights about where we can host a CI with such special CPU configuration.

Bodigrim (approved):

Lysxia (approved):

Bodigrim:
Could anyone check that the patch does not disable AVX-512 on platforms capable of it? I don't have access to modern Intel CPU at the moment.

Lysxia:
I don't have access to a AVX-512 capable CPU either.

Bodigrim:
I tried this on macOS x86_64 with AVX-512 instructions and, unfortunately, the patch just disabled their usage. I think this is a peculiarity of macOS (see https://stackoverflow.com/questions/72522885/are-the-xgetbv-and-cpuid-checks-sufficient-to-guarantee-avx2-support and https://github.com/golang/go/issues/43089), but I'm not keen to break macOS for the sake of OpenBSD. 

I guess we can guard `xgetbv` check by `#ifndef __APPLE__`, but we still need an evidence that `xgetbv` does not disable AVX-512 on Windows / Linux.

Bodigrim:
Converting to draft until someone fixes the behaviour on MacOS and tests it on Windows / Linux machines which **have** AVX-512.

falsifian:
Another problem is that this causes a SIGILL on CPUs that don't have the xgetbv instruction. [See here.](https://marc.info/?l=openbsd-ports&m=171349879216228&w=2) Checking for the osxsave CPU bit [as is done here](https://github.com/haskell/text/blob/e182733a1cc94fc716e856403f652b56b4000781/simdutf/simdutf.h#L789) and avoiding the `xgetbv` if not set would probably fix it.

(Incidentally, AVX-512 support was just added to OpenBSD-current, so hopefully the original problem no longer affects OpenBSD.)

Bodigrim:
For the record, an external workaround to disable AVX512 is to pass `-D__STDC_NO_ATOMICS__`, e. g., via `cabal.project`.

silby:
OpenBSD 7.6 shipped with AVX512 support and these patches have been dropped from the `lang/ghc` port. As I was never an expert to begin with and we don't need this in the port, I'm closing this PR, even though it seems like one could argue that something like this PR is more portable. I've probably forgotten most of what I'm learned while chasing the issue so I'm not really in a position to advise further.

PR #567, opened 2024-02-27
ci: Use haskell-actions/setup (moved from haskell/actions/setup)

Lysxia:
I saw the deprecation message in the logs. Also visible on the archived repo https://github.com/haskell/actions

### DIFF
--- a/.github/workflows/windows_and_macOS.yml
+++ b/.github/workflows/windows_and_macOS.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v3
-    - uses: haskell/actions/setup@v2
+    - uses: haskell-actions/setup@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/windows_and_macOS.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         path: ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
         key: ${{ runner.os }}-${{ matrix.ghc }}
     # We rebuild tests several times to avoid intermittent failures on Windows
-    # https://github.com/haskell/actions/issues/36
+    # https://github.com/haskell-actions/issues/36
```
Bodigrim, line 32: The new link is not valid. Maybe we do not need the workaround at all nowadays?

Bodigrim:
Thanks!

PR #571, opened 2024-03-13
Make fromString INLINEABLE

### LINKED ISSUE #549: `fromString` from `Data.Text.Lazy.Builder` generates massive amount of code
The code sample looks like

```haskell
{-# LANGUAGE OverloadedStrings #-}

f :: B.Builder
f = "a cat" <> "a dog" <> "a fish"
```

By using Builder from `bytestring`, the above code generates about 30 lines of core ([compiler explorer](https://godbolt.org/z/8KnY1bMs8)), but we got 2200 lines when switching to `text` ([compiler explorer](https://godbolt.org/z/o515q4zv8))

This becomes a problem since I use [katip](https://github.com/Soostone/katip) for logging and it uses Builder from `text` internally. It's very common to use string literal  to construct Builder while logging.

### `fromString` has INLINE

This is the most obvious reason why it produce so many code. They are inlined three times and it seems that there's nothing (eg. RULE) to make them shorter.

But it's hard for me to tell if removing `INLINE` would make performance worse or not. (And I see no benchmark involve `fromString`)

## Related issue

#19 (it's issue of pre 2.0 `text` which still use RULE to fuse though)

Lysxia:
Closes #549 

The generated Core for `f = "a cat" <> "a dog" <> "a fish" :: Data.Text.Lazy.Builder` is now more reasonable:

```
f = "a cat"#

f1 = unpackCString# f

f2 = "a dog"#

f3 = unpackCString# f2

f4 = "a fish"#

f5 = unpackCString# f4

$wf
  = \ @s x ww ww1 ww2 ww3 eta ->
      let {
        lvl
          = \ ds eta1 ->
              case ds of { Buffer ww4 ww5 ww6 ww7 ->
              $wfromString f5 x ww4 ww5 ww6 ww7 eta1
              } } in
      $wfromString
        f1
        ((\ eta1 eta2 ->
            case eta1 of { Buffer ww4 ww5 ww6 ww7 ->
            $wfromString f3 (lvl `cast` <Co:9> :: ...) ww4 ww5 ww6 ww7 eta2
            })
         `cast` <Co:9> :: ...)
        ww
        ww1
        ww2
        ww3
        eta
```

### DIFF
--- a/src/Data/Text/Internal/Builder.hs
+++ b/src/Data/Text/Internal/Builder.hs
@@ -209,7 +209,7 @@ fromString str = Builder $ \k (Buffer p0 o0 u0 l0) ->
     in loop p0 o0 u0 l0 str
   where
     chunkSize = smallChunkSize
-{-# INLINE fromString #-}
+{-# INLINEABLE fromString #-}
 
 -- | /O(1)./ A @Builder@ taking a lazy @Text@, satisfying
 --


Bodigrim (approved):

Bodigrim:
Thanks!

hasufell:
Those are @angerman's runnners. Mine should still work. You can force my runner via `runs-on: [self-hosted, Linux, ARM64, maerwald]`.

angerman:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, ... . For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

My runners don't have node-16 anymore.

hasufell:
> > Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, ... . For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
> 
> My runners don't have node-16 anymore. 

Seems you've turned a warning into an error.


> We will actively monitor the migration's progress and gather community feedback before finalizing the transition date.

Bodigrim:
Thanks, @angerman, we upgraded to Node.js 20, looks working well.

PR #572, opened 2024-03-14
Improve lazy performance of `Data.Text.Lazy.inits`

### LINKED ISSUE #562: A better `Data.Text.Lazy.inits`
#558 made me notice that while the implementation of `inits` is overall optimal ( $O(n^2)$ ), it has an awkward property that evaluating the ith text in the result takes $O(i^2)$ time. Ideally this would just take $O(i)$ time, since it involves reaching the ith element in a list and the element is a text of length i.

I believe `Data.List.inits` has solved this problem using a queue: https://hackage.haskell.org/package/base-4.19.0.0/docs/src/Data.OldList.html#inits
The same could be adopted here.

Lysxia:
Closes #562

The previous implementation, itself based on an earlier version of `Data.List.inits`, inherited the flaw that accessing the i-th element took quadratic time O(i²). This now takes linear time O(i) as expected.

The current version of `Data.List.inits` uses a banker's queue to obtain good performance when generating very long lists. For lazy text, consisting of a few big chunks, that benefit seems negligible. So I chose a simpler implementation.

Benchmarks included.

Quadratic growth before (the "2k" benchmarks take 4x the time of the "1k" benchmarks):

```
    Lazy.inits
      last 1k:      OK
        9.19 ms ± 475 μs
      last 2k:      OK
        46.7 ms ± 1.5 ms
      map-take1 1k: OK
        8.99 ms ± 798 μs
      map-take1 2k: OK
        47.0 ms ± 1.8 ms
```

Linear growth after (the "2k" benchmarks take twice the time of the "1k" benchmarks):

```
Lazy.inits
      last 1k:      OK
        46.8 μs ± 3.7 μs
      last 2k:      OK
        94.6 μs ± 1.8 μs
      map-take1 1k: OK
        66.7 μs ± 688 ns
      map-take1 2k: OK
        132  μs ± 6.2 μs
```

### DIFF
--- a/benchmarks/haskell/Benchmarks/Micro.hs
+++ b/benchmarks/haskell/Benchmarks/Micro.hs
@@ -4,19 +4,29 @@ module Benchmarks.Micro (benchmark) where
 
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text as T
-import Test.Tasty.Bench (Benchmark, bgroup, bench, nf)
+import Test.Tasty.Bench (Benchmark, Benchmarkable, bgroup, bcompareWithin, bench, nf)
 
 benchmark :: Benchmark
 benchmark = bgroup "Micro"
-  [ -- Accessing i-th element should take O(i) time.
-    -- The 2k case should run in 2x the time of the 1k case.
-    bgroup "Lazy.inits"
-      [ bench "last 1k" $ nf (last . TL.inits) (chunks 1000)
-      , bench "last 2k" $ nf (last . TL.inits) (chunks 2000)
-      , bench "map-take1 1k" $ nf (map (TL.take 1) . TL.inits) (chunks 1000)
-      , bench "map-take1 2k" $ nf (map (TL.take 1) . TL.inits) (chunks 2000)
-      ]
+  [ blinear "lazy-inits--last" 500000 2 0.1 $ \len ->
+      nf (last . TL.inits) (chunks len)
+  , blinear "lazy-inits--map-take1" 500000 2 0.1 $ \len ->
+      nf (map (TL.take 1) . TL.inits) (chunks len)
   ]
 
 chunks :: Int -> TL.Text
 chunks n = TL.fromChunks (replicate n (T.pack "a"))
+
+-- Check that running an action with input length (m * baseLen)
+-- runs m times slower than the same action with input length baseLen.
+blinear :: String  -- ^ Name (must be globally unique!)
+        -> Int     -- ^ Base length
+        -> Int     -- ^ Multiplier m
+        -> Double  -- ^ Slack s
+        -> (Int -> Benchmarkable)  -- ^ Action to measure, parameterized by input length
+        -> Benchmark
+blinear name baseLen m s run = bgroup name
+  [ bench "baseline" $ run baseLen
+  , bcompareWithin (fromIntegral m * (1 - s)) (fromIntegral m * (1 + s)) (name ++ ".baseline") $
+      bench ("x" ++ show m) $ run (m * baseLen)
+  ]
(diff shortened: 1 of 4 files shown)


src/Data/Text/Lazy.hs
```diff
@@ -1443,10 +1443,18 @@ inits = (NE.toList P.$!) . initsNE
...
-initsNE = (Empty NE.:|) . inits'
-  where inits' Empty        = []
-        inits' (Chunk t ts) = L.map (\t' -> Chunk t' Empty) (NE.tail (T.initsNE t))
-                           ++ L.map (Chunk t) (inits' ts)
+initsNE ts0 = Empty NE.:| inits' 0 ts0
+  where
+    inits' :: Int64  -- Number of previous chunks i
+           -> Text   -- The remainder after dropping i chunks from ts0
+           -> [Text] -- Prefixes longer than the first i chunks of ts0.
+    inits' _ Empty        = []
```
meooow25, line 1451: Probably should be strict in `i`?

meooow25 (approved):

benchmarks/haskell/Benchmarks/Micro.hs
```diff
@@ -0,0 +1,22 @@
...
+module Benchmarks.Micro (benchmark) where
+
+import qualified Data.Text.Lazy as TL
+import qualified Data.Text as T
+import Test.Tasty.Bench (Benchmark, bgroup, bench, nf)
+
+benchmark :: Benchmark
+benchmark = bgroup "Micro"
+  [ -- Accessing i-th element should take O(i) time.
+    -- The 2k case should run in 2x the time of the 1k case.
```
Bodigrim, line 12: Sounds like a use case for [`bcompareWithin`](https://hackage.haskell.org/package/tasty-bench-0.3.5/docs/Test-Tasty-Bench.html#v:bcompareWithin)!

src/Data/Text/Lazy.hs
```diff
@@ -1443,10 +1443,18 @@ inits = (NE.toList P.$!) . initsNE
 --
 -- @since 2.1.2
 initsNE :: Text -> NonEmpty Text
-initsNE = (Empty NE.:|) . inits'
-  where inits' Empty        = []
-        inits' (Chunk t ts) = L.map (\t' -> Chunk t' Empty) (NE.tail (T.initsNE t))
-                           ++ L.map (Chunk t) (inits' ts)
+initsNE ts0 = Empty NE.:| inits' 0 ts0
+  where
+    inits' :: Int64  -- Number of previous chunks i
```
Bodigrim, line 1448: I think number of chunks cannot exceed `Int` even on 32-bit machines.

Bodigrim (approved):
Is our test coverage good enough to generate lazy `Text` with multiple chunks?

Lysxia:
> Is our test coverage good enough to generate lazy Text with multiple chunks?

Yes, the `Arbitrary TL.Text` instance explicitly generates arbitrary chunks:

https://github.com/haskell/text/blob/240682e57dd93e793db48ea2aa98c4dfd7f89453/tests/Tests/QuickCheckUtils.hs#L98

benchmarks/haskell/Benchmarks/Micro.hs
Lysxia: Fancy!

src/Data/Text/Lazy.hs
meooow25: This is a lazy structure so it can, given enough time. For instance, `inits (cycle "text")`.

Bodigrim: Well, if we want to be precise to that extent it should be `Integer` ;)

Bodigrim:
Thanks a ton!

PR #581, opened 2024-04-13
added TextBuilder alias

BebeSparkelSparkel:


Bodigrim:
Hmm, we have [`StrictBuilder`](https://hackage.haskell.org/package/text-2.1.1/docs/Data-Text-Internal-StrictBuilder.html#v:StrictBuilder) already. Even while it's in an internal module and not widely advertised, it would be nice to come up with a consistent naming.

Should this one be `LazyBuilder` or `LazyTextBuilder` or `TextLazyBuilder`?.. @Lysxia what do you think?

Lysxia:
I like `LazyTextBuilder` since there is also [`type LazyText`](https://github.com/haskell/text/blob/master/src/Data/Text/Internal/Lazy.hs#L57), and so that leave room for `TextBuilder` as the new (and public) name for `StrictBuilder`.

BebeSparkelSparkel:
I have added `StrictTextBuilder` and `LazyTextBuilder`.

Lysxia:
I forgot we're already exporting `StrictBuilder`. So `LazyBuilder` for symmetry, instead of adding another step of renaming. Sorry for the confusion.

BebeSparkelSparkel:
I can add LazyBuilder, but I want to have Text directly in the name with StrictTextBuilder and LazyTextBuilder to differentiate it from other builder types.

Lysxia:
>  I want to have Text directly in the name with StrictTextBuilder and LazyTextBuilder to differentiate it from other builder types.

Wouldn't qualified imports let you disambiguate as well?

BebeSparkelSparkel:
I see many `type TextBuilder` aliases in packages so it seems that many people want to be able to disambiguate with the name.

https://github.com/search?q=%22type+TextBuilder%22+language%3AHaskell+&type=code

Bodigrim:
> I forgot we're already exporting `StrictBuilder`. So `LazyBuilder` for symmetry, instead of adding another step of renaming. Sorry for the confusion.

We can go for `StrictTextBuilder` / `LazyTextBuilder` and deprecate the existing `StrictBuilder` name, I don't think there are any users at the moment.

Bodigrim:
That's fine by me, but let me defer the judgement to @Lysxia as an original author of `StrictBuilder`.

Lysxia (approved):

Bodigrim (approved):

Bodigrim:
Thanks!

PR #586, opened 2024-04-27
hPutBuilder

### LINKED ISSUE #446: Add hPutBuilder for Text's Builder type
Based on discussion in #442, having a way to write a text builder directly to a handle, like [bytestring's `hPutBuilder`](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Builder.html#v:hPutBuilder) might be a nice performance win in some situations.

BebeSparkelSparkel:
closes #446 

Questions

1. If there is no encoding specified in haCodec, should should hPutBuilderUtf8 be used?
2. Does the write byte buffer actually need to be flushed?
1. Is the ByteArray in danger of being garbage collected leaving a dangerous pointer?
1. What should the buffer size be? Or, should it be a parameter?
1. With the change to Builder should `toLazyTextWith` still use `smallChunkSize`?
3. Since the buffers are lazy should there be some pre-sequencing of the next buffer. If so, how could this be done?

### DIFF
--- a/src/Data/Text/Internal/Builder.hs
+++ b/src/Data/Text/Internal/Builder.hs
@@ -40,8 +40,10 @@ module Data.Text.Internal.Builder
      -- ** The Builder type
      Builder
    , LazyTextBuilder
+   , getTexts
    , toLazyText
    , toLazyTextWith
+   , starter
 
      -- ** Constructing Builders
    , singleton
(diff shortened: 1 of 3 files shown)


Lysxia:
I don't see how you can do much better than `hPutBuilder h b = L.hPutStr h $ toLazyText b` because `Builder` has to construct a list of chunks anyway (as you mentioned yourself in the linked issue).

The only place I can see a potential speed up is if the desired encoding is UTF-8, then you could go through `Text.Utf8.hPutStr`, whose implementation could be improved by removing the `encodeUtf8` copy and writing the `Text` bytes directly to the handle.

text's `Builder`'s only purpose is to produce a `LazyText`, so I think it makes sense to keep its current API minimal and not even add `hPutBuilder`. Note that it's a quite different data structure from bytestring's `Builder`, which doesn't allocate buffers itself. Converting `Text` to bytestring's` Builder` may be a better idea than going through text's `Builder`.

PR #585, opened 2024-04-23
Added file write benchmarks

BebeSparkelSparkel:
In preparation for #446

PR #589, opened 2024-04-28
Integrate utf8 hPutStr to standard hPutStr

BebeSparkelSparkel:
`Data.Text.IO.Utf8.hPutStr` was implemented in #503 and has much better performance than `Data.Text.IO.hPutStr` when the encoding is `"UTF-8"` and the neline is `LF`, so I added it.

Question: Is there a faster way to check for the encoding without checking for string equality?


src/Data/Text/IO.hs
```diff
@@ -176,17 +179,24 @@ hGetLine = hGetLineWith T.concat
...
   case buffer_mode of
-     (NoBuffering, _)        -> hPutChars h str
-     (LineBuffering, buf)    -> writeLines h nl buf str
-     (BlockBuffering _, buf)
-         | nl == CRLF        -> writeBlocksCRLF h buf str
-         | otherwise         -> writeBlocksRaw h buf str
+    (NoBuffering, _)
+      | l > 5 && nl == LF && isUTF8   -> utf8Put
+      | otherwise                     -> hPutChars h str
+    _ | l > 80 && nl == LF && isUTF8  -> utf8Put
```
Lysxia, line 193: Why only if `l > 80`, or `l > 5` in the `NoBuffering` case?

Lysxia (commented):
Inspecting the `textEncodingName` of `TextEncoding` like you did seems like the right way.

src/Data/Text/IO.hs
BebeSparkelSparkel: Based on #585 benchmarks

Lysxia: Those magic numbers seem too fragile. They are heavily dependent on OS, hardware, and the auxiliary functions that do the actual writing. In any case I wouldn't consider `hPutStr` a performance-critical function that makes this kind of trick worth the loss in maintainability.

Branch on the simple condition without length. `utf8Put` can be optimized later if someone cares (a low hanging fruit is to avoid the copy in `encodeUtf8`).

src/Data/Text/IO.hs
```diff
@@ -176,17 +179,19 @@ hGetLine = hGetLineWith T.concat
...
-     (LineBuffering, buf)    -> writeLines h nl buf str
-     (BlockBuffering _, buf)
-         | nl == CRLF        -> writeBlocksCRLF h buf str
-         | otherwise         -> writeBlocksRaw h buf str
+    _ | nl == LF && isUTF8    -> B.hPutStr h $ encodeUtf8 t
+    (NoBuffering, _)          -> hPutChars h str
+    (LineBuffering, bufC)     -> writeLines h nl bufC str
+    (BlockBuffering _, bufC)
+        | nl == CRLF          -> writeBlocksCRLF h bufC str
+        | otherwise           -> writeBlocksRaw h bufC str
```
Lysxia, line 194: Can you revert the `buf`-to-`bufC` renaming?

src/Data/Text/IO.hs
```diff
@@ -176,17 +179,19 @@ hGetLine = hGetLineWith T.concat
...
-     (LineBuffering, buf)    -> writeLines h nl buf str
-     (BlockBuffering _, buf)
-         | nl == CRLF        -> writeBlocksCRLF h buf str
-         | otherwise         -> writeBlocksRaw h buf str
+    _ | nl == LF && isUTF8    -> B.hPutStr h $ encodeUtf8 t
+    (NoBuffering, _)          -> hPutChars h str
+    (LineBuffering, buf)     -> writeLines h nl buf str
+    (BlockBuffering _, buf)
+        | nl == CRLF          -> writeBlocksCRLF h buf str
+        | otherwise           -> writeBlocksRaw h buf str
```
Lysxia, line 194: Revert the indentation diff here. There's just one new branch here.

src/Data/Text/IO.hs
```diff
@@ -176,17 +179,19 @@ hGetLine = hGetLineWith T.concat
...
 -- This function is lifted almost verbatim from GHC.IO.Handle.Text.
 hPutStr h t = do
-  (buffer_mode, nl) <-
-       wantWritableHandle "hPutStr" h $ \h_ -> do
-                     bmode <- getSpareBuffer h_
-                     return (bmode, haOutputNL h_)
+  (buffer_mode, nl, isUTF8) <-
+    wantWritableHandle "hPutStr" h $ \h_ -> do
+      bmode <- getSpareBuffer h_
+      let isUTF8 = maybe False ((== "UTF-8") . textEncodingName) $ haCodec h_
```
Bodigrim, line 185: When is `haCodec` not `Just`? I imagine if it's `Nothing`, it means that the file is open in binary mode, so it's fine to dump UTF-8 encoded data there as well, but I did not check.

BebeSparkelSparkel: That seems like a breaking change. What would it have been before the change to utf8?

src/Data/Text/IO.hs
```diff
@@ -176,12 +179,14 @@ hGetLine = hGetLineWith T.concat
 hPutStr :: Handle -> Text -> IO ()
 -- This function is lifted almost verbatim from GHC.IO.Handle.Text.
 hPutStr h t = do
-  (buffer_mode, nl) <-
+  (buffer_mode, nl, isUTF8) <-
        wantWritableHandle "hPutStr" h $ \h_ -> do
                      bmode <- getSpareBuffer h_
-                     return (bmode, haOutputNL h_)
+                     let isUTF8 = maybe False ((== "UTF-8") . textEncodingName) $ haCodec h_
```
Bodigrim, line 185: Let's use `textEncodingName GHC.IO.Encoding.utf8` instead of hard-coded `"UTF-8"`.

Do we have any idea of performance impact? Comparing strings is not cheap. Maybe we can use `GHC.Exts.unsafePtrEquality# GHC.IO.Encoding.utf8 (haCodec h_)`. After all, it's just an optimization.

BebeSparkelSparkel: Thanks for finding this the string comparison was hurting performance a lot when I benchmarked this.

BebeSparkelSparkel: One problem is it is possible to make your own utf-8 encoding https://hackage.haskell.org/package/base-4.20.0.0/docs/GHC-IO-Encoding.html#t:TextEncoding

BebeSparkelSparkel: But who does that?

Bodigrim: If you crafted your own Unicode `TextEncoding`, you would not benefit from optimized code path, but that's fine.

It's actually safer to check `GHC.Exts.unsafePtrEquality# GHC.IO.Encoding.utf8 (haCodec h_)` than `== "UTF-8"`, because everyone can set `textEncodingName` to whatever they like.

BebeSparkelSparkel: I have tried implementing this but I am not familiar with pointer equality. If I implemented this correctly, please resolve. Otherwise, I'll try and resolve the errors.

Bodigrim (approved):

Lysxia:
I merged with master (adapting to changes from #600 which moved the logic of `hPutStr` to `Data.Text.Internal.IO.hPutStream`) (this PR is squashable, just not rebaseable) and added a comment about why the pointer-equality is fine (I initially thought it had to be guarded as ```enc `unsafePtrEquality` utf8 || enc == utf8``` but convinced myself otherwise).

Bodigrim:
@Lysxia shall we drop GHC 8.2 CI job?

src/Data/Text/Internal/IO.hs
```diff
@@ -168,17 +175,41 @@ readChunk hh@Handle__{..} buf = do
...
+
+-- | Write a string to a handle.
+hPutStr :: Handle -> Text -> IO ()
+hPutStr h t = hPutStreamOrUtf8 h (stream t) (Just putUtf8)
+  where putUtf8 = B.hPutStr h (encodeUtf8 t)
+
+-- | Write a string to a handle, followed by a newline.
+hPutStrLn :: Handle -> Text -> IO ()
+hPutStrLn h t = hPutStreamOrUtf8 h (streamLn t) (Just putUtf8)
+  where putUtf8 = hPutBuilder h (encodeUtf8Builder t <> charUtf8 '\n')
```
Bodigrim, line 188: Why not `B.hPutStrLn h (encodeUtf8 t)`?

Lysxia: It would have to be `B.hPutStrLn h (encodeUtf8 (t <> '\n'))`. I couldn't find another way to append the `'\n'` at the end in constant time. Open to suggestions.

Bodigrim: I'd expect `B.hPutStrLn` to append `'\n'` on its own, am I missing something?

Lysxia: Oh I didn't see your `Ln`. There is no `B.hPutStrLn`... right?

BebeSparkelSparkel:
```haskell
hPutStrLn :: Handle -> ByteString -> IO ()
hPutStrLn h ps = hPut h ps >> hPut h (L.singleton 0x0a)
```

https://github.com/haskell/bytestring/blob/46a3aeb179c26d2385f95ba518eaa5464c94ceaa/Data/ByteString/Lazy/Char8.hs#L932

Bodigrim: I'd say that non-100%-atomicity of `B.hPutStrLn` is for `bytestring` to resolve, not for us. But does `hPutBuilder` guarantee atomic printing?

BebeSparkelSparkel: `B.hPutBuilder` does if it can fit in the buffer.

Bodigrim: I reopened https://github.com/haskell/bytestring/issues/200 and added a comment.

Bodigrim (approved):

Bodigrim:
Thanks a ton!

PR #590, opened 2024-04-28
Fixed off by one for writeBlocksRaw

### LINKED ISSUE #587: Off by one in writeBlocksRaw
[`writeBlocksRaw`](https://github.com/haskell/text/blob/3f2a71bb1dad5fd8a64f3743b58d765d985723e2/src/Data/Text/IO.hs#L255) uses the same bounds check as [`writeBlocksCRLF`](https://github.com/haskell/text/blob/3f2a71bb1dad5fd8a64f3743b58d765d985723e2/src/Data/Text/IO.hs#L239). `writeBlocksRaw` will always only write one character while `writeBlocksCRLF` can write two.

Seems like it should actually be

```haskell
          | n >= len -> commit n True{-needs flush-} False >>= outer s
```

BebeSparkelSparkel:
closes #587

### DIFF
--- a/src/Data/Text/IO.hs
+++ b/src/Data/Text/IO.hs
@@ -252,7 +252,7 @@ writeBlocksRaw h buf0 (Stream next0 s0 _len) = outer s0 buf0
         Done -> commit n False{-no flush-} True{-release-} >> return ()
         Skip s' -> inner s' n
         Yield x s'
-          | n + 1 >= len -> commit n True{-needs flush-} False >>= outer s
+          | n >= len -> commit n True{-needs flush-} False >>= outer s
           | otherwise    -> writeCharBuf raw n x >>= inner s'
     commit = commitBuffer h raw len
 
--- a/src/Data/Text/IO.hs
+++ b/src/Data/Text/IO.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns, CPP, RecordWildCards, ScopedTypeVariables #-}
 {-# LANGUAGE Trustworthy #-}
+{-# LANGUAGE NamedFieldPuns #-}
 -- |
 -- Module      : Data.Text.IO
 -- Copyright   : (c) 2009, 2010 Bryan O'Sullivan,
@@ -54,9 +55,9 @@ import qualified Data.Text as T
 import Data.Text.Internal.Fusion (stream)
 import Data.Text.Internal.Fusion.Types (Step(..), Stream(..))
 import Data.Text.Internal.IO (hGetLineWith, readChunk)
-import GHC.IO.Buffer (Buffer(..), BufferState(..), CharBufElem, CharBuffer,
-                      RawCharBuffer, emptyBuffer, isEmptyBuffer, newCharBuffer,
-                      writeCharBuf)
+import GHC.IO.Buffer (Buffer(..), BufferState(..), RawCharBuffer, CharBuffer,
+                      emptyBuffer, isEmptyBuffer, newCharBuffer)
+import qualified GHC.IO.Buffer
 import GHC.IO.Exception (IOException(ioe_type), IOErrorType(InappropriateType))
 import GHC.IO.Handle.Internals (augmentIOError, hClose_help, wantReadableHandle,
                                 wantWritableHandle)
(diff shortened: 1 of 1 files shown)


Bodigrim:
Is it possible to add a regression test?

BebeSparkelSparkel:
It may be possible, but I am unsure how to. This is a case of the last character in the buffer not being used because the bounds check was copy pasted from the above write* functions that have the possibility of writing two characters CRLF.

Bodigrim:
Would anything in the current test suite break if we put `n - 10` instead of `n` / `n + 1`? Do we have any tests in this area at all?

BebeSparkelSparkel:
All tests pass with `n - 10`

BebeSparkelSparkel:
I don't know how to test for buffer overflow unless we can add a test buffer type that checks for that.

Lysxia:
We might not catch this in regular builds but we can test this by inserting a debug assertion (enabled by `-DASSERTS` under the `-fdeveloper` cabal flag) to ensure that `writeCharBuf` is within bounds.

src/Data/Text/IO.hs
```diff
@@ -219,42 +220,47 @@ writeLines h nl buf0 (Stream next0 s0 _len) = outer s0 buf0
           | n + 1 >= len -> commit n True{-needs flush-} False >>= outer s
           | x == '\n'    -> do
                    n' <- if nl == CRLF
-                         then do n1 <- writeCharBuf raw n '\r'
```
Lysxia, line 222: I think it looked better with `raw` before, because `buf :: Buffer` contains 4 fields that are unused in these functions. It's `commitBuffer` returning a `Buffer` that's the odd one out.

I suggest to add `len` as an argument to `writeCharBuf` or to inline the bounds check in each relevant branch.

BebeSparkelSparkel: inlined

BebeSparkelSparkel:
Still catches out of bounds https://github.com/haskell/text/actions/runs/8958991966/job/24603864107?pr=590#step:6:1102

BebeSparkelSparkel:
@Bodigrim Anything else that needs to be done?

src/Data/Text/IO.hs
```diff
@@ -206,55 +207,60 @@ hPutChars h (Stream next0 s0 _len) = loop s0
...
-  outer s1 Buffer{bufRaw=raw, bufSize=len} = inner s1 (0::Int)
+  outer s1 buf@Buffer{..} = inner s1 (0::Int)
    where
     inner !s !n =
       case next0 s of
         Done -> commit n False{-no flush-} True{-release-} >> return ()
         Skip s' -> inner s' n
         Yield x s'
-          | n + 1 >= len -> commit n True{-needs flush-} False >>= outer s
+          | n + 1 >= bufSize -> commit n True{-needs flush-} False >>= outer s
```
Bodigrim, line 220: Could this be 
```suggestion
          | n >= bufSize || x == '\n' && n >= bufSize + 1 -> commit n True{-needs flush-} False >>= outer s
```
? It seems that unless `x == '\n'` the function `writeBlocksCRLF` does nothing different from `writeBlocksRaw` and thus could use the same condition `n >= bufSize`.

Bodigrim (approved):

src/Data/Text/IO.hs
```diff
@@ -206,55 +206,48 @@ hPutChars h (Stream next0 s0 _len) = loop s0
...
-          | x == '\n'    -> do n1 <- writeCharBuf raw n '\r'
-                               writeCharBuf raw n1 '\n' >>= inner s'
-          | otherwise    -> writeCharBuf raw n x >>= inner s'
-    commit = commitBuffer h raw len
+          | n >= bufSize - bool 0 1 (isCRLF && x == '\n') ->
+            commit n True{-needs flush-} False >>= outer s
+          | isCRLF && x == '\n' -> do
+            n1 <- writeCharBuf bufRaw bufSize n '\r'
+            writeCharBuf bufRaw bufSize n1 '\n' >>= inner s'
+          | otherwise -> writeCharBuf bufRaw bufSize n x >>= inner s'
```
Lysxia, line 244: Reordering the conditions lets you mention `isCRLF` only once and makes it clearer how much buffer space the two writing branches need:

```suggestion
          | isCRLF && x == '\n' && n + 1 < bufSize -> do
              n1 <- writeCharBuf bufRaw bufSize n '\r'
              writeCharBuf bufRaw bufSize n1 '\n' >>= inner s'
          | n < bufSize -> writeCharBuf bufRaw bufSize n x >>= inner s'
          | otherwise -> commit n True{-needs flush-} False >>= outer s
```

Lysxia (approved):
LGTM!

Bodigrim:
Thanks!

Issue #442, opened 2022-06-06
`hPutStrLn` in multiple threads


worm2fed:
Function to print text with newline performs it as two separate actions - first is to print message and second - to
print newline symbol.
This way there is issues with computations in multiple threads, so newline symbol sometimes printed in wrong place.

Possible solution here is to redefine `hPutStrLn` like this:
```
hPutStrLn :: Handle -> Text -> IO ()
hPutStrLn h t = hPutStr h $ t <> '\n'
```

Not sure what is the reason of current implementation..
I found that issue when used `co-log` and I've provided PR with a fix (https://github.com/co-log/co-log/pull/243), but the source problem is `text` library.

phadej:
I'd rather document that none of `IO` actions in `Data.Text.IO` are thread-safe. Making them thread-safe would add an overhead for people who don't need one.

Bodigrim:
`bytestring` defines `hPutStrLn` as below:
```haskell
hPutStrLn :: Handle -> ByteString -> IO ()
hPutStrLn h ps
    | length ps < 1024 = hPut h (ps `B.snoc` 0x0a)
    | otherwise        = hPut h ps >> hPut h (B.singleton 0x0a) -- don't copy
```

This is still not atomic, but much less annoying in practice. I think doing the same trick in `text` + expanding documentation is a reasonable way to proceed.

axman6-da:
I initially thought _Why not use [hPutBuilder](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString-Builder.html#v:hPutBuilder) _ 
```haskell
hPutBuilder h (fromText ps <> fromText "\n")
```
but then found that the text package doesn't have an equivalent - is there a good reason for this?

worm2fed:
> `bytestring` defines `hPutStrLn` as below:
> 
> ```haskell
> hPutStrLn :: Handle -> ByteString -> IO ()
> hPutStrLn h ps
>     | length ps < 1024 = hPut h (ps `B.snoc` 0x0a)
>     | otherwise        = hPut h ps >> hPut h (B.singleton 0x0a) -- don't copy
> ```
> 

@Bodigrim but length can be expensive O(n)

axman6-da:
We don't need `Text.length` in `Char`s, we need the length of the array, which should be O(1) right?

Bodigrim:
[...]

PR #592, opened 2024-05-01
Improved hPutStr lazy for small chunk sizes

BebeSparkelSparkel:


Lysxia:
I think a simpler way is to decompose `Text.hPutStr` into `Stream.hPutStr . Text.stream` (for some `Stream.hPutStr :: Handle -> Stream Char -> IO ()` so you get `Lazy.hPutStr = Stream.hPutStr . Lazy.stream`.

BebeSparkelSparkel:
Thanks I was not aware of lazy stream.

Lysxia:
While fixing #442 I ended up doing what I suggested here, in #600.

PR #593, opened 2024-05-02
improve test function write_read

BebeSparkelSparkel:
Now tests different encodings not just UTF8.
Removed NFData constraint so that Builder can be tested.

Should be pulled before hPutStr and hPutBuilder PRs

### DIFF
--- a/tests/Tests/QuickCheckUtils.hs
+++ b/tests/Tests/QuickCheckUtils.hs
@@ -261,7 +261,7 @@ write_read unline filt writer reader nl buf ts
     IO.hSeek h IO.AbsoluteSeek 0
     r <- reader h
     let isEq = r == t
-    deepseq isEq $ pure $ counterexample (show r <> bool " /= " " == " isEq <> show t) isEq
+    deepseq isEq $ pure $ counterexample (show r ++ bool " /= " " == " isEq ++ show t) isEq
   where
     t = unline . map (filt (not . (`elem` "\r\n"))) $ ts
     encodings = [IO.utf8, IO.utf8_bom, IO.utf16, IO.utf16le, IO.utf16be, IO.utf32, IO.utf32le, IO.utf32be]
--- a/tests/Tests/QuickCheckUtils.hs
+++ b/tests/Tests/QuickCheckUtils.hs
@@ -5,6 +5,9 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
@@ -37,8 +40,11 @@ import Data.Char (isSpace)
 import Data.Text.Foreign (I8)
 import Data.Text.Lazy.Builder.RealFloat (FPFormat(..))
 import Data.Word (Word8, Word16)
-import Test.QuickCheck (Arbitrary(..), arbitraryUnicodeChar, arbitraryBoundedEnum, getUnicodeString, arbitrarySizedIntegral, shrinkIntegral, Property, ioProperty, discard, counterexample, scale, (.&&.), NonEmptyList(..), forAll)
+import GHC.IO.Encoding.Types (TextEncoding(TextEncoding,textEncodingName))
+import Test.QuickCheck (Arbitrary(..), arbitraryUnicodeChar, arbitraryBoundedEnum, getUnicodeString, arbitrarySizedIntegral, shrinkIntegral, Property, ioProperty, discard, counterexample, scale, (.&&.), NonEmptyList(..))
 import Test.QuickCheck.Gen (Gen, choose, chooseAny, elements, frequency, listOf, oneof, resize, sized)
+import Test.Tasty (TestTree, testGroup)
+import Test.Tasty.QuickCheck (testProperty)
 import Tests.Utils
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Lazy as BL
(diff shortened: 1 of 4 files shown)


tests/Tests/Utils.hs
```diff
@@ -34,8 +34,8 @@ withTempFile :: (FilePath -> Handle -> IO a) -> IO a
 withTempFile = bracket (openTempFile "." "crashy.txt") cleanupTemp . uncurry
   where
     cleanupTemp (path,h) = do
-      open <- hIsOpen h
-      when open (hClose h)
+      closed <- hIsClosed h
+      unless closed $ hClose h
```
Bodigrim, line 38: Why do we need this change? Seems equivalent to what was before?..

tests/Tests/QuickCheckUtils.hs
```diff
@@ -250,23 +250,21 @@ write_read :: (NFData a, Eq a, Show a)
            -> [a]
            -> Property
 write_read _ _ _ _ (IO.NewlineMode IO.LF IO.CRLF) _ _ = discard
-write_read unline filt writer reader nl buf ts = ioProperty $
-    (===t) <$> act
+write_read _ _ _ _ _ (IO.BlockBuffering (Just x))  _ | x < 4 = discard
+write_read unline filt writer reader nl buf ts
+  = forAll (elements encodings) $ \enc -> ioProperty $ do
```
Bodigrim, line 255: It's somewhat unfortunately that all encodings will be lumped into the same `Property`. Let's refactor this further: make `write_read` to return not a `Property` but `[TestTree]` with a test per encoding.

Bodigrim (commented):
Thanks for working on improving the test suite!

tests/Tests/Utils.hs
BebeSparkelSparkel: I was having problems with windows and the semi-closed handle state. I'll try again without.

BebeSparkelSparkel: This function will fail if the file is in a semi-closed state when doing lazy IO with LazyText and the temporary file will not be removed.

Bodigrim: Fair enough, thanks for checking.

Bodigrim (approved):

tests/Tests/QuickCheckUtils.hs
```diff
@@ -240,33 +246,36 @@ instance Arbitrary IO.BufferMode where
...
-            IO.hClose h
-            bracket (IO.openFile path IO.ReadMode) IO.hClose $ \h' -> do
-              IO.hSetEncoding h' IO.utf8
-              IO.hSetNewlineMode h' nl
-              IO.hSetBuffering h' buf
-              r <- reader h'
-              r `deepseq` return r
+  propTest :: TextEncoding -> (modeArg -> IO.BufferMode) -> (modeArg -> Bool) -> modeArg -> IO.NewlineMode -> c -> Property
+  propTest _   _               _           _      (IO.NewlineMode IO.LF IO.CRLF) _ = discard
+  propTest _   _               modeArgPred bufArg _                              _ | modeArgPred bufArg = discard
```
Lysxia, line 266: What's wrong with bigger blocks?

tests/Tests/QuickCheckUtils.hs
```diff
@@ -240,33 +246,36 @@ instance Arbitrary IO.BufferMode where
...
+  -> ((Char -> Bool) -> b -> b)
+  -> (IO.Handle -> a -> IO ())
+  -> (IO.Handle -> IO a)
+  -> (c -> [b])
+  -> [TestTree]
+write_read unline filt writer reader modData
+  = encodings <&> \enc@TextEncoding {textEncodingName} -> testGroup textEncodingName
+    [ testProperty "NoBuffering" $ propTest enc (\() -> IO.NoBuffering) (const False)
+    , testProperty "LineBuffering" $ propTest enc (\() -> IO.LineBuffering) (const False)
+    , testProperty "BlockBuffering" $ propTest enc IO.BlockBuffering (maybe False (< 4))
```
Lysxia, line 261: Rejection sampling makes the logic convoluted here. Move the sampling of `BufferMode` outside of  `propTest`.

```suggestion
    , testProperty "BlockBuffering" $ forAll (IO.BlockBuffering . fmap ((+ 4) . fromIntegral) <$> (arbitrary :: Gen Word16)) $ \blockBuffering -> propTest enc blockBuffering
```

(note the "`Gen Word16`" was copied from the `Arbitrary IO.BufferMode` instance in this file.)

tests/Tests/QuickCheckUtils.hs
```diff
@@ -240,33 +246,36 @@ instance Arbitrary IO.BufferMode where
...
+    let ts = modData d
+        t = unline . map (filt (not . (`elem` "\r\n"))) $ ts
+    IO.hSetEncoding h enc
+    IO.hSetNewlineMode h nl
+    IO.hSetBuffering h $ mkBufferingMode bufArg
+    () <- writer h t
+    IO.hSeek h IO.AbsoluteSeek 0
+    r <- reader h
+    let isEq = r == t
+    deepseq isEq $ pure $ counterexample (show r ++ bool " /= " " == " isEq ++ show t) isEq
```
Lysxia, line 277:
```suggestion
    pure (r === t)
```

`isEq` is a boolean which is about to be forced by `counterexample`, so `seq`/`deepseq` does nothing. And `===` in QuickCheck abbreviates `counterexample` of `==`.

tests/Tests/QuickCheckUtils.hs
BebeSparkelSparkel: utf 32 requires at least 4 bytes

tests/Tests/QuickCheckUtils.hs
BebeSparkelSparkel: I was having problems with lazy IO of LazyText and the handle closing before equality was tested.

I think `pure (r === t)` just passes the Property back but does not evaluate `r == t` until the handle is already closed.

tests/Tests/QuickCheckUtils.hs
BebeSparkelSparkel: moved to a generator

tests/Tests/QuickCheckUtils.hs
Lysxia: I see. Since the real culprit is lazy IO, could this be fixed at the point where `write_read` is applied to lazy IO functions (wrap them to make them strict) instead of inside `write_read`?

BebeSparkelSparkel: I think moving strictness responsibly to the caller makes the interface harder to use since it is not captured in the type signature. Also, I am planning on adding Builder to this and cannot deep sequence it but I can check for equality.

Lysxia: ok.
it should be seq though.

tests/Tests/QuickCheckUtils.hs
```diff
@@ -240,33 +249,39 @@ instance Arbitrary IO.BufferMode where
...
+    IO.hSetBuffering h mode
+    () <- writer h t
+    IO.hSeek h IO.AbsoluteSeek 0
+    r <- reader h
+    let isEq = r == t
+    deepseq isEq $ pure $ counterexample (show r ++ bool " /= " " == " isEq ++ show t) isEq
+  encodings = [IO.utf8, IO.utf8_bom, IO.utf16, IO.utf16le, IO.utf16be, IO.utf32, IO.utf32le, IO.utf32be]
+
+  blockBuffering :: Gen IO.BufferMode
+  blockBuffering = IO.BlockBuffering <$> (fmap (fmap getPositive) arbitrary) `suchThat` maybe True (> 4)
```
Lysxia, line 283: Why not apply `max 5`? Rejection sampling loops unnecessarily.

tests/Tests/QuickCheckUtils.hs
```diff
@@ -276,12 +275,12 @@ write_read unline filt writer reader modData
...
     let isEq = r == t
-    deepseq isEq $ pure $ counterexample (show r ++ bool " /= " " == " isEq ++ show t) isEq
+    seq isEq $ pure $ counterexample (show r ++ bool " /= " " == " isEq ++ show t) isEq
+
   encodings = [IO.utf8, IO.utf8_bom, IO.utf16, IO.utf16le, IO.utf16be, IO.utf32, IO.utf32le, IO.utf32be]
 
   blockBuffering :: Gen IO.BufferMode
-  blockBuffering = IO.BlockBuffering <$> (fmap (fmap getPositive) arbitrary) `suchThat` maybe True (> 4)
-
+  blockBuffering = IO.BlockBuffering <$> fmap (fmap $ min 4 . getPositive) arbitrary
```
Lysxia, line 283:
```suggestion
  blockBuffering = IO.BlockBuffering <$> fmap (fmap $ max 4 . getPositive) arbitrary
```

BebeSparkelSparkel: That was a stupid mistake thanks.

Lysxia (approved):

Bodigrim:
@BebeSparkelSparkel this is unrelated to your work here, but could you please allow `template-haskell-2.22` in the Cabal file? This would allow CI to pass.

BebeSparkelSparkel:
@Bodigrim #595

Bodigrim:
@BebeSparkelSparkel thanks a ton! Please rebase this PR.

Bodigrim:
Thanks!

PR #594, opened 2024-05-05
added to the api a placeholder for hPutBuilder

BebeSparkelSparkel:
So that hPutBuilder can be used until #586  or similar is pulled.


Bodigrim:
@BebeSparkelSparkel what's the intention here, given that #586 is closed?

PR #595, opened 2024-05-17
allow template-haskell version 2.22

BebeSparkelSparkel:


Bodigrim (approved):
Since it is trivial, let me merge without a second review.

phadej:
I'm confused. GHC-9.10.1 ships with `th-2.22` and `text-2.1.1`. The https://hackage.haskell.org/package/text-2.1.1/revisions/ was applied two days ago. 

Did GHC ship with some unversioned commit of `text`?

Bodigrim:
Shipping unversioned commit would be more or less fine, it happens all the time, but GHC was apparently shipped with a commit completely missing from upstream repo, which is quite undesirable.

phadej:
By unversioned I meant "not in the version control system" (i.e. not *untagged*)

Issue #242, opened 2018-11-25
putStrLn doesn't work correctly with LineBuffering


domenkozar:
Which can result into 

```
pushing /nix/store/j2yzfcd4wli8lj5i6yvyg6h1sbvjr55i-glibc-2.27
pushing /nix/store/7jgadbmbxp4fz4c96qm154g8sq0y8398-ghc-8.4.3-docpushing /nix/store/vf475f9cg41h8dxz5g8i0mxvbbqpry36-attr-2.4.47

pushing /nix/store/fcyxw5aslw3a82p33x5lhckj0mr6g1s4-bash-4.4-p23
```

instead of 

```
pushing /nix/store/j2yzfcd4wli8lj5i6yvyg6h1sbvjr55i-glibc-2.27
pushing /nix/store/7jgadbmbxp4fz4c96qm154g8sq0y8398-ghc-8.4.3-doc
pushing /nix/store/vf475f9cg41h8dxz5g8i0mxvbbqpry36-attr-2.4.47
pushing /nix/store/fcyxw5aslw3a82p33x5lhckj0mr6g1s4-bash-4.4-p23
```

--- 

It's currently implemented as:

```haskell
hPutStrLn :: Handle -> Text -> IO ()
hPutStrLn h t = hPutStr h t >> hPutChar h '\n'
```

The naive approach is to append the newline at the end of `t`.

GHC's hPutStrLn handles [newline as special optimization](https://github.com/ghc/ghc/blob/43967c0c7d2d0110cfc5f9d64a7dab3a3dda8953/libraries/base/GHC/IO/Handle/Text.hs#L554), I wonder if the same could be implemented for  `Data.Text`.

Bodigrim:
I don't see an issue here: `hPutStrLn` does not advertise that it works atomically.

PR #600, opened 2024-05-29
Make putStrLn more atomic with line or block buffering

### LINKED ISSUE #442: `hPutStrLn` in multiple threads

Lysxia:
Fixes #442 and #242. Subsumes #592.

This imitates the behavior in base by making the newline part of the chunk.

### DIFF
--- a/src/Data/Text/IO.hs
+++ b/src/Data/Text/IO.hs
@@ -51,7 +51,7 @@ import qualified Control.Exception as E
 import Control.Monad (liftM2, when)
 import Data.IORef (readIORef, writeIORef)
 import qualified Data.Text as T
-import Data.Text.Internal.Fusion (stream)
+import Data.Text.Internal.Fusion (stream, streamLn)
 import Data.Text.Internal.Fusion.Types (Step(..), Stream(..))
 import Data.Text.Internal.IO (hGetLineWith, readChunk)
 import GHC.IO.Buffer (Buffer(..), BufferState(..), RawCharBuffer, CharBuffer,
@@ -174,13 +174,15 @@ hGetLine = hGetLineWith T.concat
 
 -- | Write a string to a handle.
 hPutStr :: Handle -> Text -> IO ()
+hPutStr h = hPutStr' h . stream
+
 -- This function is lifted almost verbatim from GHC.IO.Handle.Text.
-hPutStr h t = do
+hPutStr' :: Handle -> Stream Char -> IO ()
+hPutStr' h str = do
   (buffer_mode, nl) <-
        wantWritableHandle "hPutStr" h $ \h_ -> do
                      bmode <- getSpareBuffer h_
                      return (bmode, haOutputNL h_)
-  let str = stream t
   case buffer_mode of
      (NoBuffering, _)        -> hPutChars h str
      (LineBuffering, buf)    -> writeLines h nl buf str
@@ -276,7 +278,7 @@ commitBuffer hdl !raw !sz !count flush release =
 
 -- | Write a string to a handle, followed by a newline.
 hPutStrLn :: Handle -> Text -> IO ()
-hPutStrLn h t = hPutStr h t >> hPutChar h '\n'
+hPutStrLn h = hPutStr' h . streamLn
 
 -- | The 'interact' function takes a function of type @Text -> Text@
 -- as its argument. The entire input from the standard input device is
--- a/src/Data/Text/IO.hs
+++ b/src/Data/Text/IO.hs
@@ -45,24 +45,18 @@ module Data.Text.IO
 import Data.Text (Text)
 import Prelude hiding (appendFile, getContents, getLine, interact,
                        putStr, putStrLn, readFile, writeFile)
-import System.IO (Handle, IOMode(..), hPutChar, openFile, stdin, stdout,
+import System.IO (Handle, IOMode(..), openFile, stdin, stdout,
                   withFile)
 import qualified Control.Exception as E
 import Control.Monad (liftM2, when)
-import Data.IORef (readIORef, writeIORef)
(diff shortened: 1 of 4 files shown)


src/Data/Text/Internal/Fusion.hs
```diff
@@ -98,6 +99,33 @@ stream (Text arr off len) = Stream next off (betweenSize (len `shiftR` 2) len)
               _ -> U8.chr4 n0 n1 n2 n3
 {-# INLINE [0] stream #-}
 
+-- | /O(n)/ @'streamLn' t = 'stream' (t <> \'\\n\')@
+streamLn ::
```
Bodigrim, line 103: I know that it's an internal module, but let's add `@since` nevertheless please.

src/Data/Text/Lazy/IO.hs
```diff
@@ -129,11 +130,11 @@ hGetLine = hGetLineWith L.fromChunks
 
 -- | Write a string to a handle.
 hPutStr :: Handle -> Text -> IO ()
-hPutStr h = mapM_ (T.hPutStr h) . L.toChunks
+hPutStr h = T.hPutStr h . stream
 
 -- | Write a string to a handle, followed by a newline.
 hPutStrLn :: Handle -> Text -> IO ()
-hPutStrLn h t = hPutStr h t >> hPutChar h '\n'
+hPutStrLn h = hPutStr h . streamLn
```
Bodigrim, line 137: Sorry, I have limited brainpower at the moment. Could we go for a more conservative alternative is to leave `Lazy.hPutStr` unchanged and implement `Lazy.hPutStrLn` by `Strict.hPutStr` for all chunks except the last, which would be printed using `Strict.hPutStrLn`?

Lysxia: Done.

src/Data/Text/Internal/Fusion.hs
Lysxia: 2.1.2 right?

Bodigrim: Yeah, 2.1.2

src/Data/Text/Lazy/IO.hs
```diff
@@ -133,7 +133,9 @@ hPutStr h = mapM_ (T.hPutStr h) . L.toChunks
 
 -- | Write a string to a handle, followed by a newline.
 hPutStrLn :: Handle -> Text -> IO ()
-hPutStrLn h t = hPutStr h t >> hPutChar h '\n'
+hPutStrLn h Empty = hPutChar h '\n'
+hPutStrLn h (Chunk t Empty) = T.hPutStrLn h t  -- print the newline after the last chunk atomically
+hPutStrLn h (Chunk t ts) = T.hPutStr t >> hPutStrLn h ts
```
BebeSparkelSparkel, line 138: > I think a simpler way is to decompose `Text.hPutStr` into `\h -> Stream.hPutStr h . Text.stream` (for some `Stream.hPutStr :: Handle -> Stream Char -> IO ()`) so you get `Lazy.hPutStr h = Stream.hPutStr h . Lazy.stream`.

`hPutStr` and `hPutStrLn` are not using the `Stream.hPutStr` mentioned in #592 . This under utilizes the character buffer for block buffering. They should be defined with `hPutStream` from Fusion.

src/Data/Text/Internal/Fusion.hs
```diff
@@ -98,6 +99,35 @@ stream (Text arr off len) = Stream next off (betweenSize (len `shiftR` 2) len)
...
+#if defined(ASSERTS)
+  HasCallStack =>
+#endif
+  Text -> Stream Char
+streamLn (Text arr off len) = Stream next off (betweenSize (len `shiftR` 2) (len + 1))
+    where
+      !end = off+len
+      next !i
+          | i > end = Done
+          | i == end  = Yield '\n' (i + 1)
```
BebeSparkelSparkel, line 115: This is almost an exact copy of `stream`. A boolean argument `appendNewline` should be added so that there is not duplication.

src/Data/Text/IO.hs
```diff
@@ -174,13 +174,15 @@ hGetLine = hGetLineWith T.concat
 
 -- | Write a string to a handle.
 hPutStr :: Handle -> Text -> IO ()
+hPutStr h = hPutStr' h . stream
+
 -- This function is lifted almost verbatim from GHC.IO.Handle.Text.
-hPutStr h t = do
+hPutStr' :: Handle -> Stream Char -> IO ()
```
BebeSparkelSparkel, line 180: Move to Fusion or Fusion.IO and name `hPutStream`

BebeSparkelSparkel (changes requested):

src/Data/Text/Lazy/IO.hs
Lysxia: Oh yeah that was before Bodigrim asked me to submit a simpler version first https://github.com/haskell/text/pull/600#discussion_r1619495047

Bodigrim (approved):

Bodigrim:
Very nice!

domenkozar:
Thank you!

PR #603, opened 2024-06-03
use hPutStream for lazy hPutStr

BebeSparkelSparkel:


Lysxia:
You need to also add `Lazy.streamLn` for `putStrLn`, to not undo #600 (previous version of #600 that did this: https://github.com/haskell/text/compare/master...Lysxia:text:atomic-putStrLn-2#diff-edae5dcf8eddff52ec5d211ae61c79a2398da37f0798e3eb96a9207f3272c2e2R135)

Bodigrim:
What's the motivation for this change? Is it for performance?

Lysxia:
Yes, for performance, by avoiding duplicate work in accessing the `Handle` for every chunk. Also more atomicity: the current implementation makes one `Strict.hPutStr` call per chunk, this PR appends chunks together into one line/block buffer before committing them. It's a continuation of #592.

Bodigrim:
> Also more atomicity: the current implementation makes one `Strict.hPutStr` call per chunk, this PR appends chunks together into one line/block buffer before committing them.

This seems to defeat the purpose of lazy `Text`, no? We do not want to append chunks together, because it would cause excessive memory consumption.

Lysxia:
The idea is that the buffer is bounded. This only happens if the chunks are smaller than the block size determined by the buffering mode.

BebeSparkelSparkel:
> We do not want to append chunks together

There is no appending of chunks; only copying to a pre-allocated character buffer.

src/Data/Text/Internal/Lazy/Fusion.hs
```diff
@@ -47,14 +48,35 @@ stream ::
   HasCallStack =>
 #endif
   Text -> Stream Char
-stream text = Stream next (text :*: 0) unknownSize
+stream = stream' False
+{-# INLINE [0] stream #-}
+
+-- | /O(n)/ @'streamLn' t = 'stream' (t <> \'\\n\')@
+--
+-- @since ????
```
Bodigrim, line 56:
```suggestion
-- @since 2.1.2
```

Bodigrim (approved):

src/Data/Text/Lazy/IO.hs
```diff
@@ -129,13 +129,11 @@ hGetLine = hGetLineWith L.fromChunks
...
 hPutStr :: Handle -> Text -> IO ()
-hPutStr h = mapM_ (T.hPutStr h) . L.toChunks
+hPutStr h = hPutStream h . stream
 
 -- | Write a string to a handle, followed by a newline.
 hPutStrLn :: Handle -> Text -> IO ()
-hPutStrLn h Empty = hPutChar h '\n'
-hPutStrLn h (Chunk t Empty) = T.hPutStrLn h t  -- print the newline after the last chunk atomically
-hPutStrLn h (Chunk t ts) = T.hPutStr h t >> hPutStrLn h ts
+hPutStrLn h t = hPutStream h $ streamLn t
```
Lysxia, line 136:
```suggestion
hPutStrLn h = hPutStream h . streamLn
```

Lysxia (approved):

Bodigrim:
Thanks!

PR #608, opened 2024-08-08
Add Data.Text.show and Data.Text.Lazy.show

### LINKED ISSUE #183: Add tshow function

Lysxia:
Closes #183

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -202,6 +202,9 @@ module Data.Text
     , zip
     , zipWith
 
+    -- * Showing values
+    , show
+
     -- -* Ordered text
     -- , sort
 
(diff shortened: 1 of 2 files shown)


Bodigrim (approved):

Bodigrim:
Awesome!

andreasabel:
Maybe this should have been in 2.2 rather 2.1.2. 
Formally it is ok to add a new function in a minor bump.
However, since `show` is in Prelude, everyone importing `Data.Text` unqualified will get breakage.
So maybe 2.2 would have been wiser...
- https://codeberg.org/valpackett/http-link-header/issues/1

Bodigrim:
It's abundantly obvious that importing `Data.Text` unqualified is unwise, because there dozens of clashes with `Prelude`, be it with `show` or without it.

Even in a hindsight, making it 2.2 would not save us any work: since `http-link-header` has no upper bounds at all, you'd still have to revise it... plus hundreds of innocent packages with `text < 2.2` atop of it.

andreasabel:
Yeah, I guess that's right.

Issue #513, opened 2023-03-24
Don't use streams and fusion


oberblastmeister:
The documentation states

> Starting from @text-1.3@ fusion is no longer implicit

However, many functions still use `Stream` internally, even thought they would probably run a lot faster without it. It is much more efficient to work on the underlying bytes instead of decoding and encoding to and from UTF-32 through `Stream`. Some functions that use `Stream` include `cons`, `snoc`, and `isPrefix`.

Bodigrim:
I'm in favor. See also https://github.com/haskell/text/issues/510#issuecomment-1439123090 for non-streaming `fold`s.

oberblastmeister:
I think folds should still use `Stream`, the issue there is that `foldl` should be defined as
```haskell
foldl = S.foldr (flip f) z . streamReverse
```

PR #611, opened 2024-08-14
Implement Data.Text.unpack directly, without streaming

Bodigrim:
Related to our discussion with @rhendric at https://discourse.haskell.org/t/the-quest-to-completely-eradicate-string-awkwardness/10111/75?u=bodigrim, let's implement `Data.Text.unpack` directly via `iterArray` instead of streaming / unstreaming. My immediate goal is to simplify dependencies between modules of `text`, but I imagine that performance could get better as well.

### DIFF
--- a/benchmarks/haskell/Benchmarks/Pure.hs
+++ b/benchmarks/haskell/Benchmarks/Pure.hs
@@ -192,6 +192,10 @@ benchmark kind ~Env{..} =
             [ benchT   $ nf T.toUpper ta
             , benchTL  $ nf TL.toUpper tla
             ]
+        , bgroup "toTitle"
+            [ benchT   $ nf T.toTitle ta
+            , benchTL  $ nf TL.toTitle tla
+            ]
         , bgroup "uncons"
             [ benchT   $ nf T.uncons ta
             , benchTL  $ nf TL.uncons tla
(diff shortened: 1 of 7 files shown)


Bodigrim:
Thanks @rhendric, should be fixed now.

rhendric:
My comment wasn't about `toTitle` specifically, but about any function with a stream-based implementation. Modify `f1` and `f2` to use any other function implemented via `unstream`, and there's still a performance regression. Example:

```hs
f1 :: String -> String
f1 = T.unpack . T.scanl max '\0' . T.pack
{-# NOINLINE f1 #-}

f2 :: String -> String
f2 = unpack . T.scanl max '\0' . T.pack
{-# NOINLINE f2 #-}
```

```
benchmarking fusion/old
time                 836.4 ns   (832.7 ns .. 841.6 ns)
mean                 833.7 ns   (831.3 ns .. 837.1 ns)
std dev              9.667 ns   (6.892 ns .. 14.57 ns)

benchmarking fusion/new
time                 1.085 μs   (1.081 μs .. 1.092 μs)
mean                 1.094 μs   (1.087 μs .. 1.104 μs)
std dev              27.77 ns   (20.48 ns .. 45.82 ns)
variance introduced by outliers: 33% (moderately inflated)
```

Bodigrim:
@rhendric we are gradually rewriting all stream-based functions to their non-streaming counterparts. E. g., see `map` and `mapAccumL`. Unless you have a particularly long pipeline, non-streaming implementation is faster. If you happen to have a long transformation pipeline (to be honest, I know exactly zero use cases for `Data.Text.scanl`), you are welcome to use `Stream` explicitly. See https://github.com/haskell/text/issues/513 for some context.

rhendric:
Ah, okay then!

Lysxia (approved):

PR #617, opened 2024-10-15
Serialise Texts without going through ByteString

alexbiehl:
It has never been easier. Ever since text is using UTF-8 encoding internally we don't have to go through the intermediate ByteString step. However, we keep deserialisation as is to ensure proper UTF-8'ness.

### DIFF
--- a/tests/Tests/Properties/Instances.hs
+++ b/tests/Tests/Properties/Instances.hs
@@ -6,6 +6,7 @@ module Tests.Properties.Instances
     ( testInstances
     ) where
 
+import Data.Binary (encode, decodeOrFail)
 import Data.String (IsString(fromString))
 import Test.QuickCheck
 import Test.Tasty (TestTree, testGroup)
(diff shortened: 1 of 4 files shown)


Bodigrim:
@alexbiehl shall we do the same for lazy `Text`?

alexbiehl:
@Bodigrim I updated the PR.

alexbiehl:
I also added property tests for strict and lazy texts.

tests/Tests/Properties/Instances.hs
```diff
@@ -43,6 +44,16 @@ tl_mempty         = mempty === (unpackS (mempty :: TL.Text))
 t_IsString        = fromString  `eqP` (T.unpack . fromString)
 tl_IsString       = fromString  `eqP` (TL.unpack . fromString)
 
+t_Binary s        =
+  case decodeOrFail . encode $ T.pack s of
```
Bodigrim, line 48: Please use `instance Arbitrary Text` from `Tests.QuickCheckUtils`, it generates more "interesting" inputs than `T.pack` does. Same below for the lazy one.

alexbiehl: Done

tests/Tests/Properties/Instances.hs
```diff
@@ -43,6 +44,16 @@ tl_mempty         = mempty === (unpackS (mempty :: TL.Text))
 t_IsString        = fromString  `eqP` (T.unpack . fromString)
 tl_IsString       = fromString  `eqP` (TL.unpack . fromString)
 
+t_Binary s        =
+  case decodeOrFail . encode $ (s :: T.Text) of
+    Left _   -> counterexample (T.unpack s) (property False)
```
Lysxia, line 49:
```suggestion
    Left _   -> counterexample (show (T.unpack s)) (property False)
```

To not dump possible control characters on the terminal

Bodigrim (approved):

Bodigrim:
Thanks!

PR #624, opened 2024-12-03
Bump dependencies (support ghc 9.12)

erikd:
`ghc-9.12` is not yet available in CI.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.19.20240514
+# version: 0.19.20250104
 #
-# REGENDATA ("0.19.20240514",["github","--config=cabal.haskell-ci","text.cabal"])
+# REGENDATA ("0.19.20250104",["github","--config=cabal.haskell-ci","text.cabal"])
 #
 name: Haskell-CI
 on:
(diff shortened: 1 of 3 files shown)


Lysxia (approved):

phadej:
> ghc-9.12 is not yet available in CI.

It is.

erikd:
> It is, for quite a while.

I was incorrectly using `9.12.0` instead of `9.12.1` :shrug:

.github/workflows/haskell-ci.yml
```diff
@@ -28,19 +28,34 @@ jobs:
     strategy:
       matrix:
         include:
+          - compiler: ghc-9.12.0.20241114
+            compilerKind: ghc
+            compilerVersion: 9.12.0.20241114
+            setup-method: ghcup
```
phadej, line 34: this is due
```
ghcup-jobs: >=8.2
```

in `haskell-ci`. `haskell-ci` tries to use the default ghcup channel for 9.12 versions too.

phadej: That configuration field should be removed. It's more harmful than useful.

Bodigrim:
I raised https://gitlab.haskell.org/ghc/ghc/-/issues/25548 to clarify whether bumping `ghc-prim` from 0.11 to 0.13 was intentional.

cabal.project
```diff
@@ -1,3 +1,8 @@
 packages: .
 tests: True
 benchmarks: True
+
+
+allow-newer:
+  , splitmix:base
+  , tasty-inspection-testing:base
```
Bodigrim, line 8: Let's keep `cabal.project` unchanged please: these `allow-newer` tend to fossilize and then no one remembers which version of GHC they are for. Besides, `tasty-inspection-testing` has been bumped already and I imagine `splitmix` will follow the suite soon.

text.cabal
```diff
@@ -53,9 +53,10 @@ tested-with:
     GHC == 9.0.2
     GHC == 9.2.8
     GHC == 9.4.8
-    GHC == 9.6.4
+    GHC == 9.6.6
     GHC == 9.8.2
```
Bodigrim, line 57: While at it, let's use 9.8.4 please.

cabal.project
erikd: I have removed the `allowed-newer` for `tasty-inspection-testing` and bracketed it with `if impl (ghc >= 9.12)`

Bodigrim:
@erikd may I ask you to update it once more? Please use the latest `haskell-ci` so that generated script uses GHC 9.12.1 proper. And there is no longer a need for `allow-newer` I believe.

Bodigrim (approved):

Bodigrim:
Thanks!

PR #627, opened 2025-03-03
Bump ghc-prim to accommodate 0.12

bgamari:
As this will be shipped with 9.12.2 per [GHC #25550](https://gitlab.haskell.org/ghc/ghc/-/issues/25550).


bgamari:
It appears that this bump has already been carried out in `master`.

PR #639, opened 2025-06-16
Fix stimes for strict text when size wraps around Int

### LINKED ISSUE #633: Bug in the new stimes size check
https://github.com/haskell/text/blob/1994b13fb873e94a5b9c492728ac000794f62f80/src/Data/Text.hs#L379-L385

It is possible to go `a` to `Int` to `a` losslessly but have a different numeric value than the `Int`. `Word` is the most common example of such a type.

So we currently get incorrect results. As an example,

```hs
  stimes (maxBound :: Word) "a"
-- ✅ fromIntegral (fromIntegral (maxBound :: Word) :: Int) == (maxBound :: Word)
= replicate (-1) "a"
= ""
```

Lysxia:
Fix #633


Bodigrim (approved):
Thanks!

PR #640, opened 2025-06-16
test: Use temporary to deal with temp files

Lysxia:
To avoid flooding the current directory with temp files when testing.

### DIFF
--- a/.github/workflows/emulated.yml
+++ b/.github/workflows/emulated.yml
@@ -30,7 +30,7 @@ jobs:
         githubToken: ${{ github.token }}
         install: |
           apt-get update -y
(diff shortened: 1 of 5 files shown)


Bodigrim:
There is also https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:emptySystemTempFile if it suits better.

Bodigrim:
https://github.com/haskell/text/blob/9f5a41a929dabd39b24af8d04b73b2b201afa5c0/.github/workflows/emulated.yml#L31-L33

Presumably this now needs also [`libghc-temporary-dev`](https://launchpad.net/ubuntu/plucky/+package/libghc-temporary-dev).

PR #649, opened 2025-07-31
Fix CRLF handling in IO functions

### LINKED ISSUE #648: CRLF with BlockBuffering not always correct
The following test case produces different results with `System.IO` and `Data.Text.IO`:

```haskell
import Control.Monad (when)
import Data.Text qualified as T
import Data.Text.IO qualified as T
import Data.ByteString.Char8 qualified as BS
import System.Directory (removeFile)
import System.IO

main :: IO ()
main = do
  let
    test putFunc value = do
      let tempFile = "test-data.txt"
      withFile tempFile WriteMode $ \h -> do
        hSetEncoding h utf8
        hSetNewlineMode h (NewlineMode CRLF CRLF)
        hSetBuffering h (BlockBuffering $ Just 8192)
        putFunc h value
      BS.readFile tempFile

    testStr = replicate 2047 '_'

  resultStr <- test hPutStrLn testStr
  resultTxt <- test T.hPutStrLn (T.pack testStr)

  when (resultStr /= resultTxt) $ do
    putStrLn "String:"
    print resultStr
    putStrLn "Text:"
    print resultTxt
```

The results are:

```text
String:
"_______________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________\r\n"
Text:
"_______________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________\n"
```

`Text` is using `"\n"` when it should be using `"\r\n"`

This doesn't happen with a different text size (eg `2046`, `2048`) and it doesn't happen with `text-2.1.1`

Lysxia:
Fix #648 

- When reading, single `\r` were read as `\n` when CRLF decoding is enabled; this bug exists since forever (16 years ago): https://github.com/haskell/text/blob/239c508db1cfdcda938b35bf39c6da2087cfcf23/Data/Text/IO.hs#L169
- When writing, `\n` was not encoded as `\r\n` when we were at the end of a buffer (bug introduced in #590, included in text-2.1.2)

Also improve the "input-output" property test to catch these mistakes (in addition to a dedicated regression test): force the use of a smaller buffer to trigger boundary conditions more easily; explicitly inspect the CRLF encoding.

And to reduce the number of test names, instead of listing all combinations of encodings and buffering modes as individual tests, they are merged into a single test that picks encodings and buffering modes randomly.

### DIFF
--- a/src/Data/Text/Internal/IO.hs
+++ b/src/Data/Text/Internal/IO.hs
@@ -112,6 +112,7 @@ unpack !buf !r !w
             | otherwise = Yield (ix i) (i+1)
     ix i = inlinePerformIO $ peekElemOff pbuf i
 
+-- Variant of 'unpack' with CRLF decoding. If there is a trailing '\r', leave it in the buffer.
 unpack_nl :: RawCharBuffer -> Int -> Int -> IO (Text, Int)
 unpack_nl !buf !r !w
  | charSize /= 4 = sizeError "unpack_nl"
@@ -130,7 +131,7 @@ unpack_nl !buf !r !w
                           in if i' < w
                              then if ix i' == '\n'
                                   then Yield '\n' (i+2)
-                                  else Yield '\n' i'
+                                  else Yield '\r' i'
                              else Done
             | otherwise = Yield c (i+1)
             where c = ix i


Bodigrim:
Thanks @Lysxia, awesome! Do you possibly have time to prepare a new `text-2.1.3` release?

Lysxia:
Yeah I can do that tomorrow

Issue #651, opened 2025-08-11
Is hGetChunk still experimental


mchav:
The Hackage documentation says this function is experimental.  Also will the API be extended to get arbitrary length chunks similar to the bytestring equivalent?

mchav:
In bytestring hGetChunk takes in an integer which is the number of bytes to read as part of the chunk. The text version only takes a handle and says it fills a buffer - it's not clear to me how big the buffer is. 

I guess the question is, and I tweak the buffer size hGetChunk reads into?

Lysxia:
Handles have a buffer of 2k `Char`s (not bytes) as defined by the internal constant `dEFAULT_CHAR_BUFFER_SIZE` in ghc-internal: https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.IO.Handle.Internals.html#dEFAULT_CHAR_BUFFER_SIZE
and AFAIK it can't be modified from any public function on `Handle`.

I also don't think it has enough utility to be worth adding to the text library.

The IO functions on Text are actually quite inefficient because they rely on `Handle` to do locale-aware decoding. As Bodigrim said, you're better off going through bytestring explicitly. If you care about performance to that extent, there are lots of design decisions you may want to tweak specifically to your application. That would be preferable to maintain outside of text.

mchav:
Thank you both. I figured with UTF 8 you might end up partially reading a character. I was asking more about the get characters. 

Bytestring to next has a conversion cost which I was trying to avoid but in light of this discussion it makes sense to pay it.

But the big take away is that this API isn't experimental anymore, right?

Bodigrim:
Yeah, not experimental anymore. A PR to update documentation is welcome :)

Bodigrim:
Documented in #655.

PR #655, opened 2025-09-19
chore: Add comprehensive documentation for `hGetChunk` (#651)

mchav:
## Summary
This PR adds detailed Haddock documentation for the `hGetChunk` function to clarify its behavior regarding UTF-8 character boundaries and buffer limitations.

## Motivation
The existing documentation didn't clearly explain why `hGetChunk` cannot guarantee reading an exact number of bytes. Users expecting byte-oriented behavior similar to `ByteString.hGetSome` may be confused by the character-oriented nature of Text operations.

## Changes
- Added comprehensive Haddock documentation covering:
  - UTF-8 character boundary handling (characters can be 1-4 bytes)
  - Internal buffer limitation (2048 characters via `dEFAULT_CHAR_BUFFER_SIZE`)
  - Clear distinction from byte-oriented operations
  - Clarifies the non-configurable buffer size limitation

### DIFF
--- a/src/Data/Text/IO.hs
+++ b/src/Data/Text/IO.hs
@@ -116,14 +116,14 @@ readChunkEof hh buf = do t <- readChunk hh buf
 -- cannot guarantee reading an exact number of bytes. Instead, it reads
 -- complete characters up to the handle's internal buffer limit.
 --
--- == Buffer Size
+-- === Buffer Size
 --
 -- The maximum chunk size is determined by the handle's internal character
 -- buffer, which is set to 2048 characters (not bytes) by the GHC runtime
 -- constant @dEFAULT_CHAR_BUFFER_SIZE@. This buffer size cannot be modified
 -- through any public API.
 --
--- == UTF-8 Considerations
+-- === UTF-8 Considerations
 --
 -- When working with UTF-8 encoded text:
 --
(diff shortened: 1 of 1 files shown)


src/Data/Text/IO.hs
```diff
@@ -101,13 +101,35 @@ readChunkEof :: Handle__ -> CharBuffer -> IO (Text, Bool)
...
+--
+-- Unlike byte-oriented functions, 'hGetChunk' operates on complete UTF-8
+-- characters. Since UTF-8 characters can occupy 1 to 4 bytes, this function
+-- cannot guarantee reading an exact number of bytes. Instead, it reads
+-- complete characters up to the handle's internal buffer limit.
+--
+-- == Buffer Size
+--
+-- The maximum chunk size is determined by the handle's internal character
+-- buffer, which is set to 2048 characters (not bytes) by the GHC runtime
```
Bodigrim, line 122: The characters for `dEFAULT_CHAR_BUFFER_SIZE` are `char` from C and each of them takes exactly 4 bytes. See https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/GHC-Internal-IO-Buffer.html#v:newCharBuffer and https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/GHC-Internal-IO-Buffer.html#v:charSize.

So I'd rather say that the default character buffer is 8192 bytes long.

Lysxia: I thought that just decoded `Char`s before writing them into the `CharBuffer` character by character, so that it doesn't matter how a `CharBuffer` represents them internally?

Bodigrim: Mm, yeah, that's probably fair.

Lysxia (approved):

Lysxia:
Thanks @mchav !

PR #659, opened 2025-09-28
Shave off redundant field of Text.Internal.Buffer

### LINKED ISSUE #654: Shave off a field of Buffer
https://github.com/haskell/text/blob/5f343f668f421bfb30cead594e52d0ac6206ff67/src/Data/Text/Internal/Builder.hs#L228-L231

Isn't the last field redundant? I imagine it can always be restored as `len(buffer) - offset - used`?..

Lysxia:
Close #654


Bodigrim (approved):

Bodigrim:
Awesome!

PR #661, opened 2025-10-10
Switch from template-haskell to template-haskell-lift

TeofilC:
We switch our dependency on template-haskell to a dependency on template-haskell-lift. This smaller library is more stabler and if we can remove the template-haskell dependency from all boot libraries then template-haskell will be much easier to re-install since it no longer needs to be in GHC's dependency closure.

For more information see the GHC proposal that introduced this library: https://github.com/ghc-proposals/ghc-proposals/pull/696

This GHC MR tests this PR against GHC-HEAD: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14978

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -227,7 +227,14 @@ library
     bytestring       >= 0.10.4 && < 0.13,
     deepseq          >= 1.1 && < 1.6,
     ghc-prim         >= 0.2 && < 0.15,
-    template-haskell >= 2.5 && < 2.25
+
+  -- template-haskell-lift was added as a boot library in GHC-9.14
+  -- once we no longer wish to backport releases to older major releases of GHC,
+  -- this conditional can be dropped
+  if impl(ghc < 9.14)
+    build-depends: template-haskell >= 2.5 && < 3
+  else
+    build-depends: template-haskell-lift >= 0.1 && <0.2
 
   if impl(ghc < 9.4)
     build-depends: data-array-byte >= 0.1 && < 0.2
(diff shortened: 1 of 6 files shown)


TeofilC:
> What's the status of this?

I'm currently waiting for GHC-9.14 to be released. And then I'll update all the PRs for boot libs.

If we want new releases of text to be backportable to older GHCs then I will need to introduce some CPP to allow building with GHC when th-lift isn't a boot lib (note that end users can build th-lift with older GHCs but we can't add new boot libs in a minor version of GHC). LMK if you'd prefer that to the current version.

Bodigrim:
> If we want new releases of text to be backportable to older GHCs then I will need to introduce some CPP to allow building with GHC when th-lift isn't a boot lib (note that end users can build th-lift with older GHCs but we can't add new boot libs in a minor version of GHC). LMK if you'd prefer that to the current version.

Yes please. We can always rip off CPP once last GHC releases in 9.10 and 9.12 series are out.

text.cabal
```diff
@@ -227,7 +227,14 @@ library
     bytestring       >= 0.10.4 && < 0.13,
     deepseq          >= 1.1 && < 1.6,
     ghc-prim         >= 0.2 && < 0.15,
-    template-haskell >= 2.5 && < 2.25
+
+  -- template-haskell-lift was added as a boot library in GHC-9.14
+  -- once we no longer wish to backport releases to older major releases of GHC,
+  -- this conditional can be dropped
+  if impl(ghc < 9.14)
+    build-depends: template-haskell
```
Bodigrim: Does it pass `cabal check` without warnings? Could we add a dummy upper bound, like `< 3`?

TeofilC: Good spot. I have restored the previous bounds

Bodigrim: Please make it `< 3`, otherwise `cabal outdated` will keep flagging it in future. (It has no notion of a dependency upper bound being implicitly guarded by GHC version)

TeofilC: I see. Done

Bodigrim:
@TeofilC could you please confirm that `text` submodule in GHC can be bumped to the current `master` (a1fd01da2bc44b255b3dd11f5a9cf098c388d6a1)?

TeofilC:
> @TeofilC could you please confirm that `text` submodule in GHC can be bumped to the current `master` ([a1fd01d](https://github.com/haskell/text/commit/a1fd01da2bc44b255b3dd11f5a9cf098c388d6a1))?

I checked and GHC compiles with the latest master of text

PR #673, opened 2025-11-26
Bump lower bound of binary to >= 0.8.3

### LINKED ISSUE #672: Dependency issues with `binary`
`text-2.1.2` and `text-2.1.3` declares that it `build-depends: binary >=0.5`, however it fails to build against `binary >=0.8.0.0 && <=0.8.2.1`:
```
src/Data/Text.hs:409:19: error:
    • Couldn't match expected type ‘binary-0.8.0.0:Data.Binary.Builder.Base.Builder’
                  with actual type ‘Data.ByteString.Builder.Internal.Builder’
      NB: ‘Data.ByteString.Builder.Internal.Builder’
            is defined in ‘Data.ByteString.Builder.Internal’
                in package ‘bytestring-0.10.8.2’
          ‘binary-0.8.0.0:Data.Binary.Builder.Base.Builder’
            is defined in ‘Data.Binary.Builder.Base’
                in package ‘binary-0.8.0.0’
    • In the first argument of ‘putBuilder’, namely
        ‘(encodeUtf8Builder t)’
      In a stmt of a 'do' block: putBuilder (encodeUtf8Builder t)
      In the expression:
        do put (lengthWord8 t)
           putBuilder (encodeUtf8Builder t)
    |
409 |       putBuilder (encodeUtf8Builder t)
    |                   ^^^^^^^^^^^^^^^^^^^
```
This is reproducible with GHC 8.4.4 and GHC 8.6.5

The reason is that starting from `binary-0.8.3.0`, `Data.Binary.Builder.Builder` became a reexport of `Data.ByteString.Builder.Builder`, where previously these were distinct types. So any code that depends on these types being the same only supports `binary >=0.8.3`.

This problem was introduced in #617

Lysxia:
Fix #672

I've also updated the bounds on Hackage for 2.1.2 and 2.1.3

### DIFF
--- a/text.cabal
+++ b/text.cabal
@@ -222,7 +222,7 @@ library
   build-depends:
     array            >= 0.3 && < 0.6,
     base             >= 4.11 && < 5,
-    binary           >= 0.5 && < 0.9,
+    binary           >= 0.8.3 && < 0.9,
     bytestring       >= 0.10.4 && < 0.13,
     deepseq          >= 1.1 && < 1.6,
     ghc-prim         >= 0.2 && < 0.15,


Bodigrim:
Thanks!

PR #675, opened 2025-12-15
Check for zero length in internal isSingleton

### LINKED ISSUE #674: `Data.Text.isSingleton` accesses out of bounds memory for zero length input
While reading some unrelated code I got distracted by this:

https://github.com/haskell/text/blob/1be3f683bba95240622f96ee1d9751a2b2df7ae9/src/Data/Text.hs#L634-L638

(introduced in https://github.com/haskell/text/commit/acf18b6d0875a28d687a692897d0865dcf52c9ad)

Now it's currently only used in places where the input can never be zero length + it's not exported.  I still see the risk that someone will use this without being aware of the unsafe nature at some point (or even worse, export it).

So maybe:
- rename this to `unsafeIsSingleton`?
- remove the Haddocks to emphasize that this is an internal function
- maybe even add a comment that states that the input must be non-zero length, but then again it's pretty obvious from the implementation, so maybe skip that

Lysxia:
Close #674

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -634,7 +634,7 @@ infixl 5 :>
 -- | /O(1)/ Tests whether a 'Text' contains exactly one character.
 isSingleton :: Text -> Bool
 isSingleton (Text arr off len) =
-  len == utf8LengthByLeader (A.unsafeIndex arr off)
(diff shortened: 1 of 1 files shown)


Bodigrim:
Thanks a lot!

PR #682, opened 2025-12-28
Fix CI

Lysxia:
OpenBSD still fails even with `-O0` and `-j1`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,7 +18,11 @@ task:
     platform: openbsd
   install_script: pkg_add ghc cabal-install git
   script:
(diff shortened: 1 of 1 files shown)


Bodigrim (approved):

Bodigrim:
I wonder if something like https://github.com/haskellari/splitmix/pull/109/files#diff-b8328e7018a96f541e9c90622b24f8306bc2ad5d2537708d200a4154374005f7R119-R123 could help with OpenBSD running out of space.

Lysxia:
Thanks for finding the fix!

PR #683, opened 2025-12-29
Fix typos

kianmeng:
Found via `codespell -L te,vew,out,yot,lamda,ot,tje,seh,otu,cche,ue,bu,yhe,writen,clen,ofo,alo,hask` and `typos --hidden --format brief`

### DIFF
--- a/simdutf/simdutf.cpp
+++ b/simdutf/simdutf.cpp
@@ -5690,7 +5690,7 @@ result base64_tail_decode(char *dst, const char_type *src, size_t length, base64
   }
 }
 
-// like base64_tail_decode, but it will not write past the end of the ouput buffer.
+// like base64_tail_decode, but it will not write past the end of the output buffer.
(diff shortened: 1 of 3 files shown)


Lysxia (approved):
OK. Consider upstreaming this to the simdutf project too.

simdutf/simdutf.cpp
```diff
@@ -6777,7 +6777,7 @@ internal::atomic_ptr<const implementation>& get_default_implementation() {
   return get_active_implementation();
 }
 #endif
-#define SIMDUTF_GET_CURRENT_IMPLEMENTION
+#define SIMDUTF_GET_CURRENT_IMPLEMENTATION
```
Bodigrim: I'm somewhat confused that this `#define` is changed only in one location. Is it unused anywhere else? 

We should probably update `simdutf` to the latest release anyway, we are quite far behind.

PR #684, opened 2025-12-29
Only run single-character case conversion tests when unicode versions match

### LINKED ISSUE #501: Guard t_toCaseFold_char with #if MIN_VERSION_base(4,16,0).
`t_toCaseFold_char` fails on `'\66937'`, which is VITHKUQI CAPITAL LETTER FE' (U+10579) added in Unicode 14.0. The test should be guarded with `#if MIN_VERSION_base(4,16,0)`.

Lysxia:
Close #501

The tests are breaking again (though with low probability) because of Unicode version mismatches between text and base. This fixes that flakiness for good by recording the unicode version and comparing it with `GHC.Unicode`.

### DIFF
--- a/tests/Tests/Properties/Text.hs
+++ b/tests/Tests/Properties/Text.hs
@@ -1,7 +1,7 @@
 -- | Tests for operations that don't fit in the other @Test.Properties.*@ modules.
 
 {-# LANGUAGE BangPatterns        #-}
-{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns        #-}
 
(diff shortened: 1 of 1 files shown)


Bodigrim (approved):

tests/Tests/Properties/Text.hs
```diff
@@ -431,6 +417,16 @@ testText =
         testProperty "ascii_toCaseFold" ascii_toCaseFold
       ],
 
+#if MIN_VERSION_base(4, 15, 0)
+      -- Requires matching version of Unicode in base and text
+      testGroup "char case conversion" $ if T.unicodeVersion == G.unicodeVersion then [
```
phadej: FWIW, it's very easy to get into state where those tests are not run at all. Any upgrade's to Unicode version used by `text` will cause these tests to not be run, as GHC is (usually) behind.

And you kind of want these tests to be run when Unicode stuff is changed...

PR #686, opened 2026-01-17
Workaround fix for inspection test broken by GHC 9.14

Lysxia:
GHC 9.14 broke inspection-testing by removing unused bindings that turn out to be used (via TH or ANN). The workaround is to trick GHC into keeping these bindings. GHC issue: https://gitlab.haskell.org/ghc/ghc/-/issues/26436


Bodigrim:
Thanks!

PR #700, opened 2026-07-10
Relax constraints on decimal, hexadecimal, hex from Integral to Num

### LINKED ISSUE #698: Relax constraints on decimal / hexadecimal from Data.Text.Read and Data.Text.Lazy.Read
`Data.Text.Read` and `Data.Text.Lazy.Read` define
```haskell
decimal :: Integral a => Reader a
hexadecimal :: Integral a => Reader a
```

I don't see a reason why they require `Integral a`, where `Num a` should suffice. I think it would be better to relax the constraint.

Lysxia:
Close #698

### DIFF
--- a/src/Data/Text/Read.hs
+++ b/src/Data/Text/Read.hs
@@ -47,7 +47,7 @@ type Parser a = IParser Text a
 -- attempt to detect overflow, so a sufficiently long input may give
 -- incorrect results.  If you are worried about overflow, use
 -- 'Integer' for your result type.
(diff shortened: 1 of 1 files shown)


Bodigrim:
My only concern is that PVP says that changes to type signatures require a major version bump. On the other hand, I see no way how this particular change can break a user.

Lysxia:
This is a low-priority change so we can leave this open, waiting for a more interesting breaking change to hitchhike with.

PR #702, opened 2026-07-10
Use memchr in indices

### LINKED ISSUE #696: Search of a singleton needle should use memchr
Surprisingly, both `Data.Text.Internal.Search` and `Data.Text.Lazy.Search` have special cases for searching a single ASCII character, but neither uses `memchr`, even while the main codepath does, so it is readily accessible.

(See https://discourse.haskell.org/t/blog-5x-speedup-changing-break-to-elemindex/13946 for context)

Lysxia:
Close #696


Bodigrim (approved):

Bodigrim:
Thanks!

PR #635, opened 2025-05-18
Change utf8LengthByLeader to a branching impl

meooow25:
The simple branching implementation is more efficient in the majority of use cases where the result is branched on anyway. In other cases, branch prediction should do a decent job on typical text.

Closes #630.

PR #705, opened 2026-07-12
Return length of 1 for continuation bytes

### LINKED ISSUE #704: Unsafe `iter` can lead to out of bounds memory access on invalid values
Hi,
the investigation of https://github.com/haskell/attoparsec/issues/232 found that the `iter` function on `Text` [[1]](https://github.com/haskell/text/blob/c0b11ac91eb1d54325c7d34b1db9612a8e8f91ee/src/Data/Text/Unsafe.hs#L71-L80) can lead to out of bounds memory access if the last byte in the text is a continuation byte and the index points to this last byte. This can happen if the last character in the text is a multibyte sequence.

This is the result of a change in 2.1.3, namely #635 which changed the branchless implementation of `utf8LengthByLeader` [[3]](https://github.com/haskell/text/blob/c0b11ac91eb1d54325c7d34b1db9612a8e8f91ee/src/Data/Text/Internal/Encoding/Utf8.hs#L85-L94) to the current branched version.
Continuation bytes in the range `0x80 - 0xBF` thus get a length of 2 bytes - this cannot be guaranteed and indeed does fail as shown above.

This happens with valid UTF-8 text. It is somewhat mitigated by the fact that the iterator is normally only applied to known valid leading bytes (usually retrieved from the output of `iter` itself) and will almost certainly return garbage anyway if not. However, the OOB issue might still be undesired.

A simple fix would be to adapt the branched implementation to also include the range `0x80 - 0xBF` in the conditional returning 1.

The relevant code snippets are appended below.
Thanks for your time,
Vekhir

### Appendix

[1] https://github.com/haskell/text/blob/c0b11ac91eb1d54325c7d34b1db9612a8e8f91ee/src/Data/Text/Unsafe.hs#L71-L80

[2] https://github.com/haskell/text/blob/c0b11ac91eb1d54325c7d34b1db9612a8e8f91ee/src/Data/Text/Unsafe.hs#L88-L98

[3] https://github.com/haskell/text/blob/c0b11ac91eb1d54325c7d34b1db9612a8e8f91ee/src/Data/Text/Internal/Encoding/Utf8.hs#L85-L94

Vekhir:
A continuation byte is always at least one byte long; returning a length of two makes accidental out of bound reads more likely. This reverts a change in behaviour in 2.1.3+ due to 1373ce315b30d24290282de8b5c8b1e8559173a0

Fixes #704

### DIFF
--- a/src/Data/Text/Internal/Encoding/Utf8.hs
+++ b/src/Data/Text/Internal/Encoding/Utf8.hs
@@ -82,13 +82,14 @@ utf8Length :: Char -> Int
 utf8Length (C# c) = I# ((1# +# geChar# c (chr# 0x80#)) +# (geChar# c (chr# 0x800#) +# geChar# c (chr# 0x10000#)))
 {-# INLINE utf8Length #-}
 
--- | Measure byte length of UTF-8 encoding for characters,
--- starting with a given byte.
+-- | Measure byte length of UTF-8 encoding for characters
+-- starting with the given leading byte.
+-- Continuation bytes yield a fail-safe value of 1.
 --
 -- @since 2.0
 utf8LengthByLeader :: Word8 -> Int
 utf8LengthByLeader w
-  | w < 0x80  = 1
+  | w < 0xC0  = 1
   | w < 0xE0  = 2
   | w < 0xF0  = 3
   | otherwise = 4


Bodigrim:
Although I don't think there is a solid technical reason for either behaviour, I don't mind merging this, because the change of behaviour in #635 was not intentional. But that assumes that we agree that the changes negate each other and do not require a major version bump in total.

Vekhir:
This isn't a breaking change in my opinion. The behaviour for continuation bytes isn't part of the interface, it's an implementation detail that may be subject to change at any time. When using the function correctly, no change can be observed.
If anything, this is a bug fix. I think it's prudent to avoid out-of-bound reads even when the function is used incorrectly.

> that assumes that we agree that the changes negate each other and do not require a major version bump in total

I do agree.

Lysxia (approved):

src/Data/Text/Internal/Encoding/Utf8.hs
```diff
@@ -88,7 +88,7 @@ utf8Length (C# c) = I# ((1# +# geChar# c (chr# 0x80#)) +# (geChar# c (chr# 0x800
 -- @since 2.0
```
Lysxia, line 89: Can you extend the doc comment to mention this behavior for continuation bytes?

Vekhir: I've updated the documentation

src/Data/Text/Internal/Encoding/Utf8.hs
```diff
@@ -82,13 +82,21 @@ utf8Length :: Char -> Int
...
+-- Continuation bytes in the range 0x80 <= w < 0xC0 are
+-- invalid leading bytes and thus have no length. To
+-- avoid out-of-bounds memory accesses, this function
+-- returns a length of 1, which is a safe dummy value
+-- as the continuation byte must be at least 1 byte long.
+--
+-- | Measure byte length of UTF-8 encoding for characters
+-- starting with the given leading byte.
+-- Continuation bytes are invalid and yield a fail-safe
+-- value of 1.
```
Lysxia:
```suggestion
-- | Measure byte length of UTF-8 encoding for characters
-- starting with the given leading byte.
-- Continuation bytes yield a fail-safe value of 1.
```

- "the continuation byte must be at least 1 byte long" is redundant (surely a byte is a byte long)
- "continuation bytes are invalid" is no longer the case now that we are explicitly deciding to give them a length.

Vekhir: Applied suggestion

PR #707, opened 2026-08-11
Add RULES for break on single character

### LINKED ISSUE #695: Add RULE from break to breakOn
`bytestring` has rewrite rules like
```haskell
{-# RULES
"ByteString specialise takeWhile (x /=)" forall x.
    takeWhile (x `neWord8`) = fst . breakByte x
"ByteString specialise takeWhile (/= x)" forall x.
    takeWhile (`neWord8` x) = fst . breakByte x
  #-}

{-# RULES
"ByteString specialise dropWhile (x /=)" forall x.
    dropWhile (x `neWord8`) = snd . breakByte x
"ByteString specialise dropWhile (/= x)" forall x.
    dropWhile (`neWord8` x) = snd . breakByte x
  #-}

{-# RULES
"ByteString specialise break (x ==)" forall x.
    break (x `eqWord8`) = breakByte x
"ByteString specialise break (== x)" forall x.
    break (`eqWord8` x) = breakByte x
  #-}
```

I think we should do the same, rewriting `Data.Text.break` as `Data.Text.breakOn` and same for lazy text.

(See https://discourse.haskell.org/t/blog-5x-speedup-changing-break-to-elemindex/13946 for context)

Lysxia:
Close #695 

We have to use `singleton` because `breakOn` expects a `Text` as the needle. Nevertheless, I checked that the rule does apply and observed a speedup in some very artificial micro-benchmarks for both the strict and lazy variant.

### DIFF
--- a/src/Data/Text.hs
+++ b/src/Data/Text.hs
@@ -228,7 +228,7 @@ import Prelude (Char, Bool(..), Int, Maybe(..), String,
                 Eq, (==), (/=), Ord(..), Ordering(..), (++),
                 Monad(..), pure, Read(..), Show,
                 (&&), (||), (+), (-), (.), ($), ($!), (>>),
-                not, return, otherwise, quot)
+                fst, not, return, otherwise, quot, snd)
 import Control.DeepSeq (NFData(rnf))
 #if defined(ASSERTS)
 import Control.Exception (assert)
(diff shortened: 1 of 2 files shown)


Bodigrim:
Thanks! Do we want the same for `takeWhile` / `dropWhile`?

Lysxia:
Sure thing. I somehow missed that we could do it with `break` (even though it's in your issue!).

Bodigrim (approved):

Bodigrim:
Thanks!

